PR #16, opened 2018-04-12
Resize observer

valdrinkoshi:
Use Resize Observer to start/stop rendering for when the container is visible.
This allows users to freely change the container size, or toggle its visibility - display: none/block - without worrying to notify the VirtualList via requestReset().

We use Resize Observer also to observe children, and remeasure when they change size.

Updated `demo/fallback.html` to show how this helps animations - changing the item size is automatically picked by the virtual-list, and repositions the other items.
animation

One nice property of ResizeObservers is that the callback is invoked even during element animations, and it seems with a 60 fps! This means that it's enough to animate a single element's `height`, and virtual-list will update all other elements' position with a 60fps - without the need of having to animate their `transform`! \o/ 
It really feels like animating a `<div>` height, which would push down the rest of the dom


valdrinkoshi:
@graynorton PTAL

demo/photos/preact/photos.js
```diff
@@ -49,7 +49,10 @@ export class Sample extends BaseSample {
           totalItems: items.length,
           component,
           resetValue,
-          ref: c => this._root = c.base
+          ref: c => {
+            this._root = c.base;
+            this._root.style.cssText = document.body.style.cssText;
```
domenic: Is this change related?

virtual-list.js
```diff
@@ -396,6 +412,27 @@ export const RepeatsAndScrolls = Superclass => class extends Repeats
     this._container.dispatchEvent(
         new RangeChangeEvent('rangechange', {first, last}));
   }
+  /**
+   * @private
+   */
+  _containerSizeChanged(size) {
+    // Include also padding.
```
domenic: I had to go look at the ResizeObserve spec to understand this comment. Maybe include more detail, like "We want to include the padding; ResizeObserve gives that to us in the left and top properties."

I am also surprised you are doing left + right instead of left + width?

valdrinkoshi: good point, will expand the comment. Weirdly enough left + width omits the padding-right, and right omits the padding-left O_o

demo/photos/preact/photos.js
valdrinkoshi: not really - this fixes some styling issue in the react photos demo (we style the body in the base class, but the react-repeater creates a `div` wrapper where it renders the content, so here we copy the body styles to the actual element that needs those) 

I Should probably move this in a separate PR...

virtual-list.js
```diff
@@ -361,20 +377,21 @@ export const RepeatsAndScrolls = Superclass => class extends Repeats
...
    */
   _shouldRender() {
-    return Boolean(
-        super._shouldRender() && this._layout && this._isContainerVisible);
+    if (!super._shouldRender() || !this._layout) {
+      return false;
+    }
+    // NOTE: we're about to render, but the ResizeObserver didn't execute yet.
+    // Since we want to keep rAF timing, we compute _containerSize now.
+    // Would be nice to have a way to flush ResizeObservers
```
valdrinkoshi: @domenic do you know of some way to force ResizeObserver to provide size?
If we wait for ResizeObserver timing, we won't render within a rAF the first time, e.g.
```js
vlist.newChild = () => document.createElement('button');
vlist.totalItems = 10;
requestAnimationFrame(() => {
  // expected to be 11 (10 buttons + 1 sizer div)
  console.log(vlist.childElementCount); 
});

demo/photos/preact/photos.js
domenic: No big deal, feel free to commit to master directly perhaps.

virtual-list.js
domenic: I think manually asking the first time seems OK, but @atotic may have ideas.

virtual-list.js
domenic: Per spec I don't actually see any way to figure out an element's padding right :(. /cc @atotic.

The spec says left = padding left, width = content width. That means automatically right = padding left + content width. So left + right = (2 * padding left) + content width. So I think this code is broken if you have uneven left/right paddings. (Or uneven top/bottom paddings.)

Maybe you just need to query the padding directly, instead of using the resize observer info? Hopefully that doesn't cause a relayout, since you'll be delivered all this stuff at a time where it's already calculated anyway?

(Although, I'm not sure why we need the padding anyway, exactly...)

valdrinkoshi: You're right, we should care of the width only

valdrinkoshi: I got initially tricked by the fact that, since we position children with `position: absolute` and set their `width: 100%`, they would get the total width of their container (padding included), so I was including that in here too... the right thing to do here is read only width/height, and if the user sets padding on the container, they'll have to include that in the children too, e.g.
```css
virtual-list {
  padding: 10px;
}
virtual-list > * {
  width: calc(100% - 20px);
}

domenic: Hmm, that's kind of annoying, maybe worth writing a separate issue to see if we can automatically make padding on the virtual-list work.

valdrinkoshi: done https://github.com/valdrinkoshi/virtual-list/issues/77

valdrinkoshi:
good to merge? @domenic @graynorton

virtual-list.js
atotic: You are correct. There is no way to get the padding from ResizeObservation. ResizeObserver wanted to support all the CSS rects initially, but could not observe them [performantly](https://github.com/WICG/ResizeObserver/issues/4), so we ended up with content rect size & position.

Workaround for observing padding of container:
    <div id="container">
        <div id="paddingObserver" style="position:absolute; top:0;left:0;right:0;bottom:0;visibility:hidden">
RO.observe(paddingObserver);

Another note:
You are calling "scheduleUpdateView" from ResizeObserver callbacks. This means you are skipping a frame because event loop ordering is:

rAF
ResizeObserver
Paint

rAF scheduled inside ResizeObserver will execute in the next event loop, after paint. If you wanted to schedule a callback before paint, Promise microtask is the way.

If you have any further question on RO use, feel free to ping me, some stuff is just non-obvious.

valdrinkoshi: Moved the discussion in the relative issues #77 and #78 👌

PR #22, opened 2018-04-17
setters for new/update/recycleChild

### LINKED ISSUE #18: Document APIs in a mdn fashion
e.g. https://developer.mozilla.org/en-US/docs/Web/API/Event/Event

### LINKED ISSUE #20: child templating: init method instead of setter?
In #13 we moved from setter-only design to constructor arguments.

In `<virtual-list>` element we chose composition over inheritance. We lazily init a `VirtualList` once the user provides the newChild method (through the `<virtual-list>.template` setter).

Rather than having the `template` setter, we could introduce an `init(config)` method where the user can pass the child templating methods, e.g.
```js
document.querySelector('virtual-list').init({
  newChild: (item, idx) => {
    const child = document.createElement('div');
    child.textContent = idx + ' - ' + item;
    return child;
  },
  // updateChild: ...,
  // recycleChild: ...,
  // direction: 'horizontal',
  // items: []
});
```

valdrinkoshi:
Fixes #18, fixes #20.

README.md made `<virtual-list>` centric, documented all APIs.

`<virtual-list>`:
- added `newChild, updateChild, recycleChild` properties
- renamed `direction` to `layout`; we already have a GridLayout implementation, so in case we decide to expose it, it makes more sense to have this property named as layout

`VirtualRepeater`:
- only `container` required in the constructor config, other properties can be set later on via setters

`VirtualList`:
- `layout` can be set via constructor args or via setter.
- renamed `stable` to `rangechange` and pass `first, last` in the `event.detail`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,30 +1,46 @@
-# Virtual list
+# &lt;virtual-list&gt;
 
-Implementation of a Virtual list API, influenced by the research done by the [infinite list study group](https://github.com/domenic/infinite-list-study-group).
+`<virtual-list>` renders only the visible portion of a list and optimizes the amount of DOM operations and reflows.
 
-## Local Setup
+## Properties
 
-Ensure you have installed the npm dependencies and serve from the project root
-```sh
-$ npm install
-$ polymer serve --npm
-```
-Then, navigate to the url: http://localhost:8081/components/virtual-list/
+- _newChild_
+  - type: `function(item: any, index: number) => Element`
+  - returns an `Element` for the given item and index.
+- _updateChild_
+  - type: `function(child: Element, item: any, index: number)`
+  - invoked after the `child` is appended to the DOM and requires to be updated with data.
+- _recycleChild_
+  - type: `function(child: Element, item: any, index: number)`
+  - invoked before the `child` is removed from the DOM. Set to control DOM removal/recycling behavior.
+- _items_
+  - type: `Array`
+  - the data model.
+- _layout_ 
+  - type: `string`
+  - layout type, can be set via attribute or property to "vertical" (default) or "horizontal".
 
-## Concepts and Domain
 
-An infinite list optimizes the rendering of DOM according to the visible area and available data.
+## Methods
 
-Some implementations append DOM incrementally, others recycle the DOM.
+- _requestReset()_
+  - Schedules a rerendering of the displayed items.
 
-## &lt;virtual-list&gt;
+## Events
 
-`<virtual-list>` can be configured through 3 properties:
-- `items (Array)`, the data model.
-- `template (Function|Object)`, generates the DOM for each data item. Can be set only once.
-- `direction (string)` (optional), layout direction, can be set via attribute or property to "vertical" (default) or "horizontal".
+- _rangechange_
+  - `bubbles: false, cancelable: false, composed: false`
+  - Fired when the list has scrolled to a new range of items.
+  - _event.detail.first_
+    - type: `number`
+    - the index of the first item currently rendered.
+  - _event.detail.num_
+    - type: `number`
+    - the number of items currently rendered.
 
-Minimal setup:
+
+## Minimal setup
+ 
 ```html
 <virtual-list></virtual-list>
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -1,30 +1,45 @@
...
-Ensure you have installed the npm dependencies and serve from the project root
-```sh
-$ npm install
-$ polymer serve --npm
-```
-Then, navigate to the url: http://localhost:8081/components/virtual-list/
+- _newChild_
+  - type: `function(item: any, index: number) => Element`
+  - returns an `Element` for the given item and index.
+- _updateChild_
```
domenic: Based on the descriptions, maybe better names for these would be childConnected and childDisconnected? I wonder if they should even be events. (But that can be a separate discussion after this PR is merged...)

README.md
```diff
@@ -1,30 +1,45 @@
...
-- `direction (string)` (optional), layout direction, can be set via attribute or property to "vertical" (default) or "horizontal".
+- _rangechange_
+  - `bubbles: false, cancelable: false, composed: false`
+  - Fired when the list has rendered to a new range of items.
+  - _event.detail.first_
+    - type: `number`
+    - the index of the first item currently rendered.
+  - _event.detail.last_
+    - type: `number`
+    - the index of the last item currently rendered.
```
domenic: This documentation is good. However, I would suggest (again it's fine to do as a follow-up) that instead of using CustomEvent, we subclass Event:

```js
class RangeChangeEvent {
  constructor(type, init) {
    super(type, init);
    this._first = init.first === undefined ? 0 : Math.floor(first);
    this._last = init.last === undefined ? 0 : Math.floor(first);
  }
  get first() {
    return this._first;
  }
  get last() {
}

File: README.md [hunk 3]
```diff
@@ -66,32 +80,47 @@ list.template = {
...
-Either set a new array to trigger the update, or use `requestReset()` to notify of changes if you want to keep the same array instance.
+`<virtual-list>` will rerender the displayed items when it receives a new `items` array.
 ```js
 list.items = list.items.concat([{name: 'new item'}]);
+```
 
+If you want to keep the same `items` array instance, use `requestReset()` to rerender the displayed items.
+```js
 list.items.push({name: 'new item'});
+list.items[0].name = 'item 0 changed!';
```
domenic: Does this version work? Does it just rerender everything currently being displayed?

README.md
```diff
@@ -1,30 +1,45 @@
...
+  - layout type, can be set via attribute or property to "vertical" (default) or "horizontal".
 
-## Concepts and Domain
 
-An infinite list optimizes the rendering of DOM according to the visible area and available data.
+## Methods
 
-Some implementations append DOM incrementally, others recycle the DOM.
+- _requestReset()_
+  - Schedules a rerendering of the displayed items.
```
domenic: We should file a follow-up issue on figuring out the best name for this (assuming we want to keep it).

virtual-repeater.js
```diff
@@ -50,7 +42,77 @@ export const Repeats = Superclass => class extends Superclass {
...
 
-    this._scheduleRender();
+    if (config) {
+      Object.assign(this, config);
+    }
+  }
+
+  // API
+
+  get container() {
```
domenic: Seems like there's a few things not documented here. Worth filing an issue to keep track of them so we can decide whether to document them or make them internal.

README.md
valdrinkoshi: tracking issue: https://github.com/PolymerLabs/virtual-list/issues/26

README.md
valdrinkoshi: Nice idea, will include it in this PR

README.md
valdrinkoshi: Yes, it will rerender all the displayed items. If the user does recycling this translates to a bunch of array pop()/push() + updateChild()

virtual-repeater.js
valdrinkoshi: Yeah, these components need some documentation indeed. I wanted to decide first how these will be shipped, in order to focus efforts where is needed first.
Tracking issue https://github.com/PolymerLabs/virtual-list/issues/27

PR #30, opened 2018-04-24
Flesh out README

domenic:
The content is very similar, just expanded a bit. This should be more like a web platform explainer.

Please review for accuracy; I might have gotten things wrong while explaining them more!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,93 +1,142 @@
 # &lt;virtual-list&gt;
 
-`<virtual-list>` renders only the visible portion of a list and optimizes the amount of DOM operations and reflows.
+`<virtual-list>` maps a provided set of JavaScript objects onto DOM nodes, and renders only the DOM nodes that are currently visible, leaving the rest "virtualized".
 
-## Properties
+This document is an early-stage explainer for `<virtual-list>` as a potential future web platform feature, as part of the [layered API](https://github.com/drufball/layered-apis) project. The repository also hosts a proof-of-concept implementation that is being co-evolved with the design.
 
-- _newChild_
-  - type: `function(item: any, index: number) => Element`
-  - returns an `Element` for the given item and index.
-- _updateChild_
-  - type: `function(child: Element, item: any, index: number)`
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,93 +1,142 @@
...
 
-`<virtual-list>` will rerender the displayed items when it receives a new `items` array.
 ```js
 list.items = list.items.concat([{name: 'new item'}]);
 ```
 
-If you want to keep the same `items` array instance, ensure to update the DOM via `updateChild` and to use `requestReset()` to rerender the displayed items.
+In this case, `newChild` will be called for every item, including the ones that already had corresponding elements in the old items array.
+
+If you want to keep the same `items` array instance, you can use `requestReset()` to rerender the currently-displayed items. If you do this, you'll also need to set `updateChild`, since the elements will no longer be created from scratch. For example:
```
valdrinkoshi: Probably here we can be more explicit on when to use `updateChild`: the user should always set `updateChild` if they plan to update the array contents, specifically if the dom was already created for that index. e.g.:
```js
list.items = ['item'];
requestAnimationFrame(() => {
  // The list has already created a child for index = 0
  // so the user needs to define updateChild in order 
  // to update that DOM node.
  // The new item (index = 1) will be created via newChild
  list.items = ['changed!', 'new item'];
});
```
In short, it's just safer to update the DOM with `updateChild` if you plan to manipulate data.
Maybe it makes sense to mention this at the top of this section?

README.md
```diff
@@ -1,93 +1,142 @@
...
+
+### `recycleChild` property
+
+Type: `function(child: Element, item: any, index: number)`
+
+Set this property to replace the default behavior of removing an item's element from the DOM when it is no longer visible.
+
+This is often used for node-recycling scenarios, as seen in [the example below](#dom-recycling).
+
+_We are discussing the naming and API for this functionality in [#25](https://github.com/PolymerLabs/virtual-list/issues/25)._
```
valdrinkoshi: Links here and below need to be updated to the new repo

domenic:
I think I clarified it; let me know!

PR #36, opened 2018-04-27
expose itemKey

### LINKED ISSUE #32: [feature] expose itemKey function
The `Repeats` mixin keeps track of the created DOM in the key/child map `_keyToChild`. When a child is not found, it will invoke `newChild` to create it and stores it in the `_keyToChild` map.
In this example `newChild` is invoked only for the newly added item, while `updateChild` is invoked for all the items:
```js
list.items = list.items.concat({name: 'new item'});
```

`_keyToChild` uses index as the key. This can be customized by providing the `itemKey` function to return a custom key `itemKey(item: any) => string|number`

Should we expose this property in `<virtual-list>`?

valdrinkoshi:
Fixes #32 

Added demo showing how using `itemKey` can affect rendering and animation
sorting-contacts

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,22 +160,54 @@ In this case, `newChild` will be called for the newly-added item once it becomes
 
 ### Efficient re-ordering
 
-`<virtual-list>` keeps track of the generated DOM via an internal key/Element map, in order to limit the number of created nodes via `newChild`. The default key is the array index.
+`<virtual-list>` keeps track of the generated DOM via an internal key/Element map to limit the number of created nodes. 
 
-When manipulating data, you might be re-ordering items. For example, imagine we have a contact list and the following method to move a contact to the end:
+The default key is the array index, but can be customized through the `itemKey` property.
+
+Imagine we have a list of 3 contacts:
 ```js
-const moveContactToEnd = (contact) => {
-  myContacts.splice(myContacts.indexOf(contact), 1); // remove it
+const myContacts = [{name: 'A'}, {name: 'B'}, {name: 'C'}];
+virtualList.items = myContacts;
+virtualList.newChild = () => document.createElement('div');
+virtualList.updateChild = (div, contact) => div.textContent = contact.name;
+```
+This renders 3 contacts, and the `<virtual-list>` key/Element map is:
+```
+0: <div>A</div>
+1: <div>B</div>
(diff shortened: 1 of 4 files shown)


valdrinkoshi:
@domenic PTAL, in particular the explainer part (README.md)

@graynorton FYI

README.md
```diff
@@ -150,6 +158,57 @@ list.requestReset();
...
+```
+This renders 3 contacts, and the `<virtual-list>` key/Element map is:
+```
+0: <div>A</div>
+1: <div>B</div>
+2: <div>C</div>
+```
+We want to move the first contact to the end:
+```js
+const moveFirstContactToEnd = () => {
```
domenic: Nit: I would suggest using normal `function` syntax here.

README.md
```diff
@@ -63,6 +63,14 @@ This is often used for node-recycling scenarios, as seen in [the example below](
 
 _We are discussing the naming and API for this functionality in [#25](https://github.com/valdrinkoshi/virtual-list/issues/25)._
 
+### `itemKey` property
+
+Type: `function(child: Element) => string|number`
```
domenic: string|number is strange. We should state how this is actually sorted. What if I mix strings and numbers? Does this just use JS's `array.sort()`?

valdrinkoshi: `itemKey` doesn't affect the sorting directly, as it just provides an identifier for a given item. `Repeats` uses this identifier to figure out if we should generate the DOM or not (if to invoke `newChild`)

domenic: Oh, right, that makes sense.

So why does it have to be a string or number? Couldn't it be anything?

valdrinkoshi: ...indeed! Will update the signature here :)

valdrinkoshi: What's an efficient way to say that it should be defined in the signature?
` => any` makes it look like it can be null or undefined

domenic: I think `=> any` is correct. If you return `null` or `undefined` that's bad in the same way that always returning `1` is bad.

Eventually we'll need to formally specify what happens if you do bad things, but for now we can just document the happy path.

valdrinkoshi: sg. Also, I realized that the input was incorrect, the function doesn't receive `Element` but the data itself...Corrected!

PR #42, opened 2018-05-01
<virtual-list> as default scrollTarget

### LINKED ISSUE #23: [feature] set scrolling element

valdrinkoshi:
Fixes #23 by making `<virtual-list>` the default scroll target instead of the main document.
`<virtual-list>` now needs to be explicitly sized in order to render the items. The default size is 300 x 150 pixels, same as for `<iframe>` and ``

We also reflect `layout` value to attribute so that we can conditionally set `max-width: 100%` for `layout = 'vertical'` and `max-height: 100%` for `layout = 'horizontal'`. Without this, it would not be possible to have children bigger than the `<virtual-list>`, e.g.
```html
<style>
  virtual-list { height: 200px }
  virtual-list > div { height: 400px }
</style>
<virtual-list id="list"></virtual-list>
<script>
 list.items = new Array(10);
 list.newChild = () => document.createElement('div');
 list.updateChild = (child) => child.textContent = 'tall div';
</script>
```


README.md
```diff
@@ -227,6 +227,27 @@ list.addEventListener('rangechange', (event) => {
...
+  /* https://www.w3.org/TR/CSS2/visudet.html#inline-replaced-height */
+  height: 150px;
+}
+```
+
+Main document scrolling will be achievable through [`document.rootScroller`](https://www.chromestatus.com/feature/4533350404456448)
+```html
+<virtual-list style="height: 100vh"></virtual-list>
+<script type="module">
+  import './node_modules/virtual-list/virtual-list-element.js';
```
domenic: Most of our other examples don't include this import; the one that does uses `std:virtual-list|https://some.cdn.com/virtual-list.js`. I think we should leave it out.

README.md
```diff
@@ -227,6 +227,27 @@ list.addEventListener('rangechange', (event) => {
...
+```css
+virtual-list {
+  /* https://www.w3.org/TR/CSS2/visudet.html#inline-replaced-width */
+  width: 300px;
+  /* https://www.w3.org/TR/CSS2/visudet.html#inline-replaced-height */
+  height: 150px;
+}
+```
+
+Main document scrolling will be achievable through [`document.rootScroller`](https://www.chromestatus.com/feature/4533350404456448)
```
domenic: We should link to https://github.com/bokand/root-scroller

README.md
```diff
@@ -227,6 +227,27 @@ list.addEventListener('rangechange', (event) => {
 });
 ```
 
+### Scrolling
+
+`<virtual-list>` needs to be sized in order to determine how many items should be rendered. Its default size is:
```
domenic: I think showing the default size with CSS is not that helpful. I'd instead just say "300px × 150px, similar to [CSS inline replaced elements](https://www.w3.org/TR/CSS2/visudet.html#inline-replaced-width) like images and iframes"

PR #43, opened 2018-05-02
Use VirtualList instead of <virtual-list>

### LINKED ISSUE #19: html-spec-viewer breaks styles
The html spec stylesheet styles elements with selectors like `body > p`, hence hosting the content inside `<html-spec-viewer>` breaks these styles.

valdrinkoshi:
Fixes #19 by using VirtualList instead of the custom element, so that `body` selectors can work


domenic:
Hmm, but now this is no longer a demo of the thing that we're proposing to standardize :(. I think I'd rather have some broken styles but still demo the right thing...

In other words I think the proper fix for #19 is just noting that because we're hacking together a demo from an existing page, there will be some minor imperfections, which wouldn't be there in a page built from scratch.

valdrinkoshi:
Oh ok, I should have just marked #19 as a wontfix. Will revert this PR

PR #49, opened 2018-05-03
demo/fallback.html

### LINKED ISSUE #44: demo/fallback.html
Prepare demo to showcase a fallback UI while loading items

valdrinkoshi:
Fixes #44 by implemeting an example with fallback UI elements
Moved `<contact-element>` into its own file, and added a `sortable` attribute which will display render count and up/down buttons.

##  fallback.html demo

fallback

## bugfixes

While implementing the demo, I found and fixed 2 bugs:
- the value of `last` provided in `rangechange` was offset of -1 
- there might be a race condition when measuring children: measurement is done 1 microtask after `_render()`, and `_render()` is throttled of a microtask, so it might be that 2 consecutive `_render()` might affect the measurement (e.g. the 2nd render hides some children, and the 1st _measureChildren would memoize a 0x0 rect)
screen shot 2018-05-04 at 11 59 37 am
^ to fix this, we debounce _measureCallback

### DIFF
--- a/demo/fallback.html
+++ b/demo/fallback.html
@@ -71,7 +71,7 @@
 
     // Start with fallbacks, generate enough to cover the viewport height.
     const fallbacksCount = Math.ceil(virtualList.offsetHeight / 80);
-    virtualList.items = getFallbacks(fallbacksCount);
+    virtualList.items = createFallbacks(fallbacksCount);
 
     // Keep track of first fallback index to know when we need to load more contacts.
     let firstFallbackIndex = 0;
@@ -97,7 +97,7 @@
       virtualList.requestReset();
     }
 
-    function getFallbacks(count) {
+    function createFallbacks(count) {
       const fallbacks = [];
       for (let i = 0; i < count; i++) {
         fallbacks.push({
--- a/demo/fallback.html
+++ b/demo/fallback.html
@@ -69,24 +69,28 @@
       element.contact = contact;
     };
 
-    // Start with fallbacks, generate enough to cover the viewport height.
+    // Generate enough fallbacks to cover the viewport height.
     const fallbacksCount = Math.ceil(virtualList.offsetHeight / 80);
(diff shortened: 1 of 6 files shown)


valdrinkoshi:
@graynorton @domenic PTAL

demo/fallback.html
```diff
@@ -0,0 +1,152 @@
...
+  </p>
+
+  <virtual-list></virtual-list>
+
+  <script type="module">
+    import '../virtual-list-element.js';
+    import './contacts/contact-element.js';
+
+    const fallbackHeight = 80;
+    const fallbacksCount = Math.ceil(innerHeight / fallbackHeight) + 5;
```
domenic: Is this window.innerHeight? Maybe being explicit is better?

A comment explaining why this is calculated this way would be nice.

valdrinkoshi:
Good points, updated! PTAL

virtual-repeater.js
```diff
@@ -222,8 +224,17 @@ export const Repeats = Superclass => class extends Superclass {
    */
   async _measureChildren() {
     if (this._ordered.length > 0) {
+      // Grab the children to be measured. We do this
+      // synchronously as _toMeasure relies on _prevFirst and _prevLast.
       const {indices, children} = this._toMeasure;
+      // We debounce the measurement of a microtask.
+      const id = ++this._measuringId;
       await Promise.resolve();
```
valdrinkoshi: @graynorton do you have suggestions on how to "justify" this microtask?
I was thinking this was added to give time to classes/styles to be applied, but by this time we already did call _newChild and _updateChild, so the user has had already the time to update styles. Also, getBoundingClientRect() forces relayout.

valdrinkoshi: Actually, this helps play nice with browser timing and avoids forcing layout, while still being executed before the next frame. Here a comparison of the demo with & without microtask wait:

 sync execution | await microtask 
---| ---
no-wait |  wait


So, we keep it!

PR #55, opened 2018-05-04
Remove default width

### LINKED ISSUE #53: remove default width: 300px
`<virtual-list>` can leverage the default behavior of `display: block` which will take all the available width.
http://output.jsbin.com/beqigu/1/quiet
http://jsbin.com/beqigu/1/edit?html,output

valdrinkoshi:
Fixes #53 by removing the default width and leveraging the `display: block` behavior.
We still require a default height.

Also, set width/height instead of max-width/max-height for distributed content, as we expect the children to occupy the full width of a vertical list, and full height of a horizontal list. For grid lists we expect each child to be sized.

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,7 +285,7 @@ list.addEventListener('rangechange', (event) => {
 
 ### Scrolling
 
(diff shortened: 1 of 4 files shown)


valdrinkoshi:
That change is for the distributed content - the virtual-list children - in particular we set width: 100% only when layout = vertical, and height: 100% when layout = horizontal

domenic:
Ah right, my eyes skimmed over the `::slotted(*)`. Looks good then!!

PR #56, opened 2018-05-04
rename itemKey to childKey

### LINKED ISSUE #52: rename itemKey to childKey
`itemKey(item: any) => any` is used to determine an identifier for the DOM corresponding to the array item.
We should probably rename it to `childKey`, as it is effectively a sort of "id" for the stamped dom.

We could name it `childId`, but that might lead users to think that the generated dom has an `id` attribute.

valdrinkoshi:
Fixes #52

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,11 +68,11 @@ This is often used for node-recycling scenarios, as seen in [the example below](
 
 _We are discussing the naming and API for this functionality in [#25](https://github.com/valdrinkoshi/virtual-list/issues/25)._
 
-### `itemKey` property
+### `childKey` property
 
 Type: `function(item: any) => any`
 
-Set this property to provide an identifier for a given item.
+Set this property to provide a custom identifier for the element corresponding to a given item.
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -68,11 +68,11 @@ This is often used for node-recycling scenarios, as seen in [the example below](
...
 
-### `itemKey` property
+### `childKey` property
 
 Type: `function(item: any) => any`
 
-Set this property to provide an identifier for a given item.
+Set this property to provide a custom identifier for the element corresponding to a given item.
 
 This is often used for more efficient re-ordering, as seen in [the example below](#efficient-re-ordering-using-itemkey).
```
domenic: The anchor will change so you'll need to change this link too.

valdrinkoshi: great catch, updated in https://github.com/valdrinkoshi/virtual-list/pull/56/commits/eb4930177bc9592619fea6ee939647dad5b6bb2f

PR #57, opened 2018-05-05
restore dom, let lit-html handle the update

### LINKED ISSUE #39: Rendering issue with https://valdrinkoshi.github.io/virtual-list/demo/contacts/lit-html/
Click on one of the items (in Chrome) and change the text with the keyboard... then scroll, like moving the scrollbar with the mouse - lots of rendering artifacts. At one point I had the same item stay visible always while scrolling

valdrinkoshi:
Fixes #39


valdrinkoshi:
@graynorton PTAL

PR #59, opened 2018-05-07
demo/header-footer.html

### LINKED ISSUE #45: demo/header-footer.html
Demo showcasing a virtual-list with headers/footers

valdrinkoshi:
Fixes #45

### DIFF
--- a/demo/header-footer.html
+++ b/demo/header-footer.html
@@ -35,15 +35,29 @@
 
     const virtualList = document.querySelector('virtual-list');
 
+    const nodePool = { group: [], contact: [] };
     virtualList.newChild = (item) => {
+      let element;
       if (item.group) {
-        const groupEl = document.createElement('header');
-        groupEl.textContent = `${item.title} (${item.count})`;
-        return groupEl;
+        element = nodePool.group.pop() || document.createElement('header');
+      } else {
+        element = nodePool.contact.pop() || document.createElement('contact-element');
       }
-      const contactEl = document.createElement('contact-element');
-      contactEl.contact = item;
-      return contactEl;
+      return element;
+    };
+    virtualList.updateChild = (element, item) => {
+      if (item.group) {
+        element.textContent = `${item.title} (${item.count})`;
+      } else {
+        element.contact = item;
+      }
+    };
+    virtualList.recycleChild = (element, item) => {
+      if (item.group) {
+        nodePool.group.push(element);
+      } else {
+        nodePool.contact.push(element);
+      }
     };
 
     // Get contacts, sort and group by name.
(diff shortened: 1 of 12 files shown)


demo/header-footer.html
```diff
@@ -0,0 +1,99 @@
...
+
+  <virtual-list></virtual-list>
+
+  <script type="module">
+    import '../virtual-list-element.js';
+    import './contacts/contact-element.js';
+
+    const virtualList = document.querySelector('virtual-list');
+
+    const nodePool = { group: [], contact: [] };
```
domenic: Add a comment here to explain that our items come in two shapes: contacts, or groups = `{ group: true, title, count }`. So, each of the following functions treats these cases different, and we have two different node pools for recycling among them.

demo/header-footer.html
```diff
@@ -0,0 +1,99 @@
...
+    virtualList.recycleChild = (element, item) => {
+      if (item.group) {
+        nodePool.group.push(element);
+      } else {
+        nodePool.contact.push(element);
+      }
+    };
+
+    // Get contacts, sort and group by name.
+    fetch('contacts/contacts.json')
```
domenic: Expand the comment here to explain what the input looks like and what the output will look like.

demo/header-footer.html
```diff
@@ -0,0 +1,99 @@
...
+      .then((resp) => resp.json())
+      .then((contacts) => {
+        let group = null;
+        return contacts.sort((a, b) => {
+          const nameA = a.name.toUpperCase(); // ignore upper and lowercase
```
domenic: You can just do `(a, b) => a.name.localCompare(b.name)`.

demo/header-footer.html
```diff
@@ -0,0 +1,99 @@
...
+          if (nameA < nameB) {
+            return -1;
+          }
+          if (nameA > nameB) {
+            return 1;
+          }
+          // names must be equal
+          return 0;
+        }).reduce((items, item) => {
+          let title = item.name.substr(0, 1);
```
domenic: Better is `item.name[0]`.

demo/header-footer.html
```diff
@@ -0,0 +1,99 @@
...
+          const nameB = b.name.toUpperCase(); // ignore upper and lowercase
+          if (nameA < nameB) {
+            return -1;
+          }
+          if (nameA > nameB) {
+            return 1;
+          }
+          // names must be equal
+          return 0;
+        }).reduce((items, item) => {
```
domenic: Maybe `item` should be named `contact`?

demo/header-footer.html
domenic: It might be better to use async-await for this chain

valdrinkoshi:
Nice tips @domenic, demo got even clearer 👌

PR #62, opened 2018-05-07
contacts/lit-html/ ensure start/end marker comment placement 

### LINKED ISSUE #39: Rendering issue with https://valdrinkoshi.github.io/virtual-list/demo/contacts/lit-html/

valdrinkoshi:
Fixes #39 by ensuring that there are no text nodes in `<p contenteditable>` and that we always place start/end marker comments before committing the change


valdrinkoshi:
LGTMed offline by @graynorton \o/

PR #63, opened 2018-05-07
use std:virtual-list for <html-spec-viewer>

valdrinkoshi:
Add html-spec-native-lapi/ which loads the built-in virtual-list. Added instructions on screen when flag is not enabled:
screen shot 2018-05-07 at 4 20 27 pm

Updated demo page as follows:
screen shot 2018-05-07 at 3 50 34 pm

### DIFF
--- a/demo/html-spec-native-lapi/index.html
+++ b/demo/html-spec-native-lapi/index.html
@@ -15,8 +15,8 @@
     <a href="https://www.chromestatus.com/feature/5466425791610880">Streams API: TransformStream</a> support.
   </p>
 
-  <div id="lapiRequirement">This demo uses the built-in virtual-list from the Layered API. Enable one of the following flags in chrome://flags to try
-    it out:
+  <div id="lapiRequirement">This demo uses the built-in virtual-list from the Layered API. If you are using Chrome, ensure you are using version 68.0.3420.0
+    or greater, and enable one of the following flags in chrome://flags to try it out:
     <ul>
       <li>#enable-layered-api</li>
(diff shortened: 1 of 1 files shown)


demo/html-spec-native-lapi/index.html
```diff
@@ -0,0 +1,39 @@
...
+
+<body>
+
+  <html-spec-viewer></html-spec-viewer>
+
+  <p id="streamRequirement">This demo requires
+    <a href="https://www.chromestatus.com/feature/5466425791610880">Streams API: TransformStream</a> support.
+  </p>
+
+  <div id="lapiRequirement">This demo uses the built-in virtual-list from the Layered API. Enable one of the following flags in chrome://flags to try
```
domenic: I'd say: "If you are using Chrome, ensure you are using version 68.0.3420.0 or greater, and enable one..."

Other browsers might have different instructions.

demo/html-spec-native-lapi/index.html
```diff
@@ -0,0 +1,39 @@
...
+      <li>#enable-experimental-web-platform-features</li>
+    </ul>
+  </div>
+
+  <script type="module">
+    if (window.TransformStream) {
+      document.getElementById('streamRequirement').remove();
+      import('./html-spec-viewer.js')
+
+      customElements.whenDefined('html-spec-viewer').then(() => {
```
domenic: You should be able to just do `import('./html-spec-viewer.js').then(() => {` instead of these two steps, I think.

Issue #29, opened 2018-04-24
Type of items property


domenic:
Right now the `items` setter just takes whatever its given and stores it. The getter returns the stored value.

The more web-platformey thing to do would be:

- The setter accepts any iterable, snapshotting that iterable at the time it's set
- The getter returns a frozen array of the snapshot

The main benefit here is that you can't write code like

```js
list.items.push(...);
```

because the array is frozen. (Such code won't work, because there's no notification that the array changed. You'd instead need to use `requestReset()`: see #26.)

The drawback of this is that it's O(n) in the total number of items. The current requestReset() pattern allows O(1) item additions and removals.

Not sure what to do here, but I wanted to have a tracking issue open.

valdrinkoshi:
Closing this as now we expose `totalItems: number` instead of `items: array` - see #65

PR #65, opened 2018-05-10
totalItems: number

### LINKED ISSUE #21: items (array) vs totalItems (number)
Neither VirtualRepeater nor VirtualList actively use `items` or mark in any way the objects in the array, so it should be possible to setup a list with only `totalItems` (number) instead of `items` (array). This would translate in a setup like this:
```js
const items = new Array(10).fill('item');

/* old setup with items (array) */
new VirtualRepeater({
  items: items,
  newChild: (item, idx) => document.createElement('div'),
  updateChild: (child, item, idx) => child.textContent = `${idx} - ${item}`,
  childKey: (item) => item.key
});

/* new setup with totalItems (number) */
new VirtualRepeater({
  totalItems: items.length,
  newChild: (idx) => document.createElement('div'),
  updateChild: (child, idx) => child.textContent = `${idx} - ${items[idx]}`,
  childKey: (idx) => items[idx].key
});
```

One drawback of this approach is that we cannot check anymore for identity in the `totalItems` setter, meaning that we'd have to always reset the list when `totalItems` is set, whereas with `items` we could check for strict identity in order to trigger the reset.

valdrinkoshi:
Fixes #21, #29.

This PR removes the `items: array` property in favor of the `totalItems: number` property. As a consequence, we don't pass the `item` as input for new/update/recycleChild and childKey, but rather the index of the item.

This exposes even more the need for `requestReset()` as a mean to notify the `virtual-list` that the data has changed.

One nice property of this change is that now the user could theoretically set `virtualList.size = Number.MAX_SAFE_INTEGER` and dynamically load each item as it gets displayed, without the need to allocate an array of that size.
```js
const nodePool = [];
virtualList.newChild = () => nodePool.pop() || document.createElement('div');
virtualList.updateChild = (elem, i) => elem.textContent = `item ${i}`;
virtualList.recycleChild = (elem) => nodePool.push(elem);
// Render as many items as possible --> infinite scrolling!
virtualList.totalItems = Number.MAX_SAFE_INTEGER;
```
In practice, we'd hit the maximum scroll height handled by the browser. In Chrome this seems to be ~33.5M pixels 📜

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ The (tentative) API design choices made here, as well as the list's capabilities
 
   // This will automatically cause a render of the visible children
   // (i.e., those that fit on the screen).
-  list.items = myItems.length;
+  list.size = myItems.length;
 </script>
 ```
 
@@ -49,12 +49,12 @@ This property is required. Without it set, nothing will render.
 
 Type: `function(child: Element, itemIndex: number)`
 
-Set this property to configure the virtual list with a function that will update items' elements.
+Set this property to configure the virtual list with a function that will update the element with data at a given index.
 
 If set, this property is invoked in two scenarios:
 
 * The user scrolls the list, changing which items' elements are visible. In this case, `updateChild` is called for all of the newly-visible elements.
-* The developer sets the `items` property, changing the contents of an item whose element is visible.
+* The developer changes the `size` property.
 * The developer calls `requestReset()`, which will call `updateChild` for all currently-visible elements. See [below](#data-manipulation-using-requestreset) for why this can be useful.
 
 For more on the interplay between `newChild` and `updateChild`, and when each is appropriate, see [the example below](#using-newchild-and-updatechild)
@@ -73,17 +73,17 @@ _We are discussing the naming and API for this functionality in [#25](https://gi
 
 Type: `function(itemIndex: number) => any`
 
-Set this property to provide a custom identifier for the element corresponding to a given item.
+Set this property to provide a custom identifier for the element corresponding to a given data index.
 
 This is often used for more efficient re-ordering, as seen in [the example below](#efficient-re-ordering-using-childkey).
 
-### `items` property
+### `size` property
 
 Type: `number`
 
-Set this property to control how many items are contained in the list. (The items are mapped to elements via the `newChild` and `updateChild` properties.)
+Set this property to control how many items the list will display. The items are mapped to elements via the `newChild` property, so this controls the total number of times `newChild` could be called, as the user scrolls to reveal all the times.
 
-Can also be set as an attribute on the element, e.g. `<virtual-list items="10"></virtual-list>`
+Can also be set as an attribute on the element, e.g. `<virtual-list size="10"></virtual-list>`
 
 ### `layout` property
 
@@ -100,9 +100,9 @@ Can also be set as an attribute on the element, e.g. `<virtual-list layout="hori
 
 ### `requestReset()` method
 
-This re-renders all of the currently-displayed items, updating them from their source data using `updateChild`.
+This re-renders all of the currently-displayed elements, updating them from their source data using `updateChild`.
 
-This can be useful when you mutate data without changing the amount of `items`. Also see [the example below](#data-manipulation-using-requestreset).
+This can be useful when you mutate data without changing the `size`. Also see [the example below](#data-manipulation-using-requestreset).
 
 _We are discussing the naming of this API, as well as whether it should exist at all, in [#26](https://github.com/valdrinkoshi/virtual-list/issues/26). The aforementioned [#29](https://github.com/valdrinkoshi/virtual-list/issues/29) is also relevant._
 
@@ -138,7 +138,7 @@ list.newChild = index => {
 };
 
 // Calls newChild four times (assuming the screen is big enough)
-list.items = myItems.length;
+list.size = myItems.length;
 ```
 
 In this example, we are statically displaying a virtual list with four items, which we never plan to update. This can be useful for use cases where you would otherwise use static HTML, but want to get the performance benefits of virtualization. (Admittedly, we'd need more than four items to see that happen in reality.)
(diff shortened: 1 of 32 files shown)


README.md
```diff
@@ -38,15 +39,15 @@ Checkout more examples in [demo/index.html](./demo/index.html).
...
+Set this property to configure the virtual list with a factory that creates an element the first time a given item at the specified index is ready to be displayed in the DOM.
 
 This property is required. Without it set, nothing will render.
 
 ### `updateChild` property
 
-Type: `function(child: Element, item: any, index: number)`
+Type: `function(child: Element, itemIndex: number)`
 
 Set this property to configure the virtual list with a function that will update items' elements.
```
domenic: This documentation doesn't seem quite updated. This sentence should probably say something like "the list's elements" or "the element at a given index"? And, the second bullet point seems like it should probably just be deleted.

README.md
```diff
@@ -70,19 +71,19 @@ _We are discussing the naming and API for this functionality in [#25](https://gi
 
 ### `childKey` property
 
-Type: `function(item: any) => any`
+Type: `function(itemIndex: number) => any`
 
 Set this property to provide a custom identifier for the element corresponding to a given item.
```
domenic: Maybe "element corresponding to a given index"?

README.md
```diff
@@ -70,19 +71,19 @@ _We are discussing the naming and API for this functionality in [#25](https://gi
...
 ### `childKey` property
 
-Type: `function(item: any) => any`
+Type: `function(itemIndex: number) => any`
 
 Set this property to provide a custom identifier for the element corresponding to a given item.
 
 This is often used for more efficient re-ordering, as seen in [the example below](#efficient-re-ordering-using-childkey).
 
 ### `items` property
```
domenic: We might change this soon anyway, but I'd find `totalItems` to be a better name probably.

README.md
```diff
@@ -70,19 +71,19 @@ _We are discussing the naming and API for this functionality in [#25](https://gi
...
 
-Type: `Array`
+Type: `number`
 
-Set this property to control the items which are displayed in the list. (The items are mapped to elements via the `newChild` and `updateChild` properties.)
+Set this property to control how many items are contained in the list. (The items are mapped to elements via the `newChild` and `updateChild` properties.)
```
domenic: This probably needs more explanation now. My attempt:

> Set this property to control how many items the list will display. The items are mapped to elements via the `newChild` property, so this controls the total number of times `newChild` could be called, as the user scrolls to reveal all the times.

README.md
```diff
@@ -70,19 +71,19 @@ _We are discussing the naming and API for this functionality in [#25](https://gi
...
 
-_Right now the getter for this property just returns back the set value. We are discussing how exactly that should work in [#29](https://github.com/valdrinkoshi/virtual-list/issues/29)._
+Can also be set as an attribute on the element, e.g. `<virtual-list items="10"></virtual-list>`
```
domenic: Nice bonus

README.md
```diff
@@ -101,7 +102,7 @@ Can also be set as an attribute on the element, e.g. `<virtual-list layout="hori
 
 This re-renders all of the currently-displayed items, updating them from their source data using `updateChild`.
 
-This can be useful for if you mutate the `items` array, or elements in it, instead of setting the `items` property to a new value. Also see [the example below](#data-manipulation-using-requestreset).
+This can be useful when you mutate data without changing the amount of `items`. Also see [the example below](#data-manipulation-using-requestreset).
```
domenic: We'll probably want a bigger documentation and naming overhaul for requestReset(), since it's now very very important. But we can do that as a follow-up; I'm eager to help.

README.md
```diff
@@ -154,38 +158,41 @@ _Note: see [#15](https://github.com/valdrinkoshi/virtual-list/issues/51) for why
 If you plan to update your items, you're likely better off using `newChild` to set up the "template" for each item, and using `updateChild` to fill in the data. Like so:
 
 ```js
-list.newChild = item => {
+list.newChild = (index) => {
```
domenic: Maybe just change this to `()` while here, actually.

README.md
```diff
@@ -195,26 +202,26 @@ Object.assign(list, {
 
 ### Data manipulation using `requestReset()`
 
-The `<virtual-list>` element will automatically rerender the displayed items when it receives a new `items` array. For example, to add a new item to the end, you could do:
+The `<virtual-list>` element will automatically rerender the displayed items when `items` changes. For example, to add a new item to the end, you could do:
 
 ```js
-list.items = list.items.concat([{name: 'new item'}]);
+myItems.push('new item');
+list.items++;
```
domenic: Hmm, having to write that kind of code seems a bit annoying. Really easy to forget the `list.items++`. I hope the data source ideas we talked about make it better.

virtual-list-element.js
```diff
@@ -63,12 +63,14 @@ export class VirtualListElement extends HTMLElement {
...
 
   static get observedAttributes() {
-    return ['layout'];
+    return ['layout', 'items'];
   }
 
   attributeChangedCallback(name, oldVal, newVal) {
     if (name === 'layout') {
       this.layout = newVal;
+    } else if (name === 'items') {
```
domenic: Don't do this pattern for reflecting attributes. Instead, define

```js
get size() {
  return +this.getAttribute("size");
}

set size(v) {
  this.setAttribute("size", +v);
}
```

and then use `this.size` internally. This handles several edge cases better.

README.md
valdrinkoshi: Will change to 
> Set this property to configure the virtual list with a function that will update the element **with data** at a given index.

Re second bullet point, we should still mention that `updateChild` is invoked if `size` is changed

virtual-list-element.js
valdrinkoshi: For `size` we don't require to reflect it to attribute. We do that for `layout` because we have need it for styling.
Also, `size` must be a number, so it's more performant to do the conversion only in the `attributeChangedCallback` and expect users to pass in a number when using the setter.

domenic: It's not really OK to have a property that sometimes picks up its value from an attribute, but doesn't reflect back. That's not something we see in native DOM elements. They all use the pattern I showed above.

In general you cannot expect users to pass in the right types in DOM APIs. They all do conversions.

valdrinkoshi: Ok, made all reflecting and always reading from the attribute now

PR #67, opened 2018-05-14
rAF batching

### LINKED ISSUE #51: Let's discuss microtask-batched reading of items array
Per https://github.com/valdrinkoshi/virtual-list/pull/48#discussion_r185677707, apparently virtual-list only reads from the items array after a microtask delay. This is pretty unusual, and I think we should discuss it.

In particular, the example discussed there is pretty confusing. Consider a list where you have only set `newChild`. If you do

```js
// (1)
list.items = x;
list.items = y;
```

then it will call `newChild` for all the items in y. But if you do

```js
// (2)
list.items = x;
setTimeout(() => {
  list.items = y;
}, 0);
```

Then it will call `newChild` for all the items in x, and the items in y are ignored. I would prefer that both are consistent: i.e. both do newChild for all items in x.

Now consider a list where you have set both `newChild` and `updateChild`. Then (1) will call newChild + updateChild for all the items in y, and (2) will call newChild for all the items in x, and updateChild for all the items in y. I would prefer that both are consistent: i.e. both do newChild for all items in x, and updateChild for all items in y.

---

Note that the platform has some precedent for delayed rendering. `img.src = x; img.src = y;` will only try to load y. This case feels worse because of the newChild/updateChild distinction, though.

MutationObservers are also microtask-batched, but they don't do coalescing, so they tell you *all* the things that happened since the last time, not just the latest thing.

---

I'm open to being convinced that batching is more important than consistency. But I lean toward saying that if developers change `items` a lot, they should know that will cause a lot of updates, and we shouldn't try to batch for them.

Maybe if developers want batching, we should have something more suited to their use case, e.g.

```js
const resumeUpdates = list.suspendUpdates();
... mutate list.items a lot ...
resumeUpdates();
```

Also, if we do think batching is important, I think microtask batching is probably not right, and we should do requestAnimationFrame batching instead.

Thoughts?

valdrinkoshi:
Fixes #51 by changing the internal implementation to batch all updates around a single rAF.

We schedule render updates with `_scheduleRender()`.

In `_render()`, we update the viewport and scroll position of the layout, and we keep rendering and reflowing the layout until we reach stability.

Note that Layout notifies of range changes and children position changes synchronously, but VirtualList actually needs to render the new range and then it can position the dom. For this reason, we keep track of the `_childrenPos` and then schedule a render where we actually position the children.

We expose a `reflow()` function on Layout so that we can synchronously render in VirtualList.

### DIFF
--- a/layouts/layout-1d-base.js
+++ b/layouts/layout-1d-base.js
@@ -197,7 +197,14 @@ export default class Layout extends EventTarget {
   }
 
   _scheduleReflow() {
-    this._pendingReflow = true;
+    if (!this._pendingReflow) {
+      this._pendingReflow = Promise.resolve().then(() => {
+        if (this._pendingReflow) {
+          this._pendingReflow = null;
+          this._reflow();
+        }
+      });
+    }
   }
 
   _updateScrollSize() {
--- a/layouts/layout-1d-base.js
+++ b/layouts/layout-1d-base.js
@@ -170,7 +170,12 @@ export default class Layout extends EventTarget {
     // Override
   }
 
-  // TODO: Does this need to be public?
+  reflow() {
+    if (this._pendingReflow) {
+      this._pendingReflow = null;
+      this._reflow();
+    }
+  }
   _reflow() {
     const {_first, _last, _scrollSize} = this;
 
@@ -189,17 +194,10 @@ export default class Layout extends EventTarget {
       this._emitRange();
       this._emitChildPositions();
     }
-    this._pendingReflow = null;
   }
 
   _scheduleReflow() {
-    if (!this._pendingReflow) {
-      this._pendingReflow = Promise.resolve().then(() => {
-        if (this._pendingReflow) {
-          this._reflow();
-        }
-      });
-    }
+    this._pendingReflow = true;
   }
 
   _updateScrollSize() {
--- a/layouts/layout-1d-base.js
+++ b/layouts/layout-1d-base.js
@@ -194,7 +194,11 @@ export default class Layout extends EventTarget {
 
   _scheduleReflow() {
     if (!this._pendingReflow) {
-      this._pendingReflow = Promise.resolve().then(() => this._reflow());
+      this._pendingReflow = Promise.resolve().then(() => {
+        if (this._pendingReflow) {
+          this._reflow();
+        }
+      });
     }
   }
 
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -5,7 +5,7 @@ This document gives an overview of various pieces we use to build up the `<virtu
 ## VirtualRepeater (Repeats mixin)
 
 - Orchestrates DOM creation and layouting, ensures minimum number of nodes is created.
-- Given an `items` array, it displays `num` elements starting from `first` index.
+- Given a `totalItems` amount, it displays `num` elements starting from `first` index.
 - Delegates DOM creation, update and recycling via `newChild, updateChild, recycleChild`.
 - Delegates DOM layout via `_measureCallback`.
 
@@ -14,9 +14,9 @@ This document gives an overview of various pieces we use to build up the `<virtu
 ```js
 const repeater = new VirtualRepeater({
   /**
-   * The data model.
+   * Total number of items.
    */
-  items: new Array(20).fill({name: 'item'}),
+  totalItems: myItems.length,
   /**
    * From which index to start.
    */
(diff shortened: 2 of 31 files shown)


valdrinkoshi:
@graynorton PTAL

layouts/layout-1d-base.js
```diff
@@ -189,12 +188,19 @@ export default class Layout extends EventTarget {
...
       this._emitChildPositions();
     }
-    this._pendingReflow = null;
   }
 
   _scheduleReflow() {
     if (!this._pendingReflow) {
-      this._pendingReflow = Promise.resolve().then(() => this._reflow());
+      this._pendingReflow =
+          Promise.resolve().then(() => this.flushPendingReflow());
```
domenic: This looks suspicious; why is it done? I think all these kinds of things need an extensive comment explaining why they're delayed instead of synchronous...

valdrinkoshi: Uhm, I guess we can now avoid doing batching here. That removes the need of `flushPendingReflow`.

The reason this is here is so that it doesn't matter which order we set layout properties like `direction, itemSize, viewportSize`. I'll look more into it, and at minimum add a comment there 👍

valdrinkoshi: So I played with this for a while, and if we make Layout reflow synchronously when a property changes, it becomes very important in which order properties are set. Also, reflow can be computationally expensive, so it doesn't make sense to trigger 3 reflows when the user sets 3 properties.

For coherency, I would keep microtask batching here and offer the `flushPendingReflow` method to force sync reflow.

In `<virtual-list>` I've removed the microtask batching, but I should probably put it back, for the same reasons.

domenic: I am not very concerned with the performance when someone changes `layout`; that should be rare. I think the predictability of having only one batching strategy is more important.

I want to stress again that microtask batching is just not acceptable; it does not match anything on the platform. requestAnimationFrame based batching is the only realistic option.

An alternative you may want to consider is that, since all of this is delayed until requestAnimationFrame time anyway, you can reorder the changes as necessary to get best performance. But do so using normal code, not using microtask scheduling.

valdrinkoshi: Then what if we don't auto-reflow on property change, but just mark that we need to reflow?
The user will have to invoke `reflow()` themselves to apply the updates.

If yes, should we do the same for VirtualList? properties won't trigger render, but the user will have to invoke it themselves (in our case this will be done by `<virtual-list>`)

valdrinkoshi: Something like this:
```js
const layout = new Layout();
layout.updateItemSizes({...});
layout.reflow();

const list = new VirtualList();
list.totalItems = 10;
list.newChild = (i) => ...
list.layout = layout;
list.container = ...
list.render();

domenic: I'm confused which user you are talking about. In general a strategy of marking what's needed so the next time the rAF callback comes around you do the reflow makes sense. But I don't want to require people to do `listElement.reflow()`.

valdrinkoshi: The `<virtual-list>` user wouldn't require to do `listElement.reflow()`, because `<virtual-list>` would handle internally rendering & reflow in its `[_render]()` function.

We have demoes that use `VirtualList` and `Layout` instances instead of `<virtual-list>` e.g. demo/basic/, demo/contacts/, demo/photos/

It feels weird for the `VirtualList/Layout` user that changing a property on `VirtualList` triggers dom updates, but changing properties on `Layout` does not.

The `<virtual-list>` user doesn't have access to the `Layout` instance, at least not yet.

valdrinkoshi: talked offline with @domenic, and agreed to have only Layout "passive" for now - in the sense that the user needs to call reflowIfNeeded() to trigger reflow.
Like this, the timing of the updates can be more predictable.

valdrinkoshi: Updated implementation and PR description 👌

valdrinkoshi:
In this PR we rely on the fact that we can do things synchronously inside the render loop (measure children, force layout reflow). This allows us to paint only once before the next frame.

If we use ResizeObserver to retrieve children sizes, or if Layout is inside a Web Worker, we might have to do at least 2 render loops which would span in 2 frames.

#### single frame 
single-frame

#### multi frame
I just removed the [`if (!this._pendingRender)` check in the render loop](https://github.com/valdrinkoshi/virtual-list/pull/67/files#diff-470077c6f5a1416593e10b09f1becd2eR197)
multi-frame

PR #75, opened 2018-05-17
placeholders for html-spec-viewer

valdrinkoshi:
spec-viewer

We now check if an item is still attached before removing it. This allows implementors to reparent a child without worrying to notify the VirtualScroller about it. Specifically in html-spec-viewer, this allows to create only 3 placeholders, even if during the first rendering the Layout requires to generate more than 3 items (because its default itemSize is 150px).


valdrinkoshi:
I'm looking if I can load the items at the scrolled index & below - that would be awesome!

valdrinkoshi:
Will merge this as is, unfortunately jumping to a specific item requires anyways parsing the whole spec, so this can't be easily done (my hope was to jump to a specific index without parsing) .

PR #84, opened 2018-05-29
*Child -> *Element

### LINKED ISSUE #25: Provide better explanation for updateChild and recycleChild
`newChild` function is invoked when the `<virtual-list>` needs DOM to represent an item.

`recycleChild` function is invoked when the `<virtual-list>` has already set `display: none` on the child and is about to remove it from the document. if not defined, `<virtual-list>` will remove the child from the document. The user can use this property to stop the default behavior and manage the DOM.
When defined, `<virtual-list>` will not remove the child from the document to avoid a removeChild/appendChild sequence for performance reasons. The user can still decide to remove the child from the document in the `recycleChild` function.

`updateChild` function is invoked after the `<virtual-list>` has appended the child to the container and requires it to be updated with data.

1) Is there a better name that conveys the intent of these functions?
2) shoulde we fire an event instead of `recycleChild`? Something like `willremovechild`? We'd still need a way to know that it was handled (e.g. via `preventDefault()`) in order to prevent the removal of the child from the document.
3) ~~do we really need `updateChild`? the user could just use `newChild` and `recycleChild`~~ 
```js
const vList = document.querySelector('virtual-list');
const pool = [];
Object.assign(vList, {
  items: new Array(20).fill('item'),
  newChild: (item, idx) => {
    const child = pool.pop() || document.createElement('div');
    child.textContent = idx + ' - ' + item;
    return child;
  },
  recycleChild: (child, item, idx) => {
    pool.push(child);
  }
});
```
^ We need `updateChild` for cases like data updates w/o changing `first, num`

valdrinkoshi:
Fixes #25.
Renamed methods `new/update/recycleChild` to `create/update/recycleElement`, and `childKey` to `elementKey`.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -6,7 +6,7 @@ This document gives an overview of various pieces we use to build up the `<virtu
 
 - Orchestrates DOM creation and layouting, ensures minimum number of nodes is created.
 - Given a `totalItems` amount, it displays `num` elements starting from `first` index.
-- Delegates DOM creation, update and recycling via `createElement, updateChild, recycleChild`.
+- Delegates DOM creation, update and recycling via `createElement, updateElement, recycleChild`.
 - Delegates DOM layout via `_measureCallback`.
 
 ### Basic setup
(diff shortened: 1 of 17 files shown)


valdrinkoshi:
Reviewed & approved offline by @domenic  👌

PR #86, opened 2018-06-05
Default create/update/recycleElement

### LINKED ISSUE #80: default values for newChild and updateChild
What about setting these default values for newChild and updateChild?
```js
constructor() {
  super();
  this.newChild = () => document.createElement('div');
  // assuming we have itemsSource API
  this.updateChild = (child, item) => child.innerHTML = item.toString();
}
```
The user can still override these values to customize the scroller.

A default `newChild` helps especially those template libraries that need a container to render within, e.g. lit-html and preact. With lit-html, the user would just need to do this to update the nodes:
```js
import {render, html} from './node_modules/lit-html/lib/lit-extended.js';
import './virtual-scroller-element.js';

const itemTemplate = (item) => html`item ${item.someproperty}`;
const updateChild = (child, item) => render(itemTemplate(item), child);

const scrollerTemplate = (totalItems) => html`
   <virtual-scroller updateChild=${updateChild} totalItems=${totalItems}>
   </virtual-scroller>`;
      
render(scrollerTemplate(1000), document.body);

valdrinkoshi:
Fixes #80.
With these new defaults, the user only needs to set `updateElement` to effectively customize the rendering. 
`virtual-scroller` creates a `<div>` as a child template, but this can be customized by distributing a `<template>` with any element into the scroller. The scroller will recycle these children by default.

Updated README.md and couple demos as an example of usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,9 +58,7 @@ This property is invoked in these scenarios:
 * The developer changes the `totalItems` property.
 * The developer calls `requestReset()`, which will call `updateElement` for all currently-visible elements. See [below](#data-manipulation-using-requestreset) for why this can be useful.
 
-The default `updateElement` sets the textContent of the child to be the data index.
-
-For more on the interplay between `createElement` and `updateElement`, and when each is appropriate, see [the example below](#using-createelement-and-updateelement)
+The default `updateElement` sets the textContent of the child to be the data index. Almost all uses of `<virtual-scroller>` will want to change this behavior.
 
 ### `recycleElement` property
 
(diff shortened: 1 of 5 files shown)


domenic:
I pushed more readme updates. One thing I was thinking is that resetting recycleElement should not reset createElement. It would be nice to be able to turn off recycling while still getting default element creation. What do you think?

valdrinkoshi:
Uhm, yeah that should be possible. So:
- if user sets a custom `createElement`, we should set `recycleElement` to null (if it's still the default).
- if the sets a custom `recycleElement`, we should just empty the nodePool, so that the default `createElement` will always create new children.

domenic:
Yep, that sounds perfect to me.

valdrinkoshi:
Done 👌 There is a chance for this to be a shotgun for the user, for example if the user sets `recycleElement !== null` while keeping the default `createElement`, they'll be effectively keeping those elements in the dom while creating new elements. But I feel like that's an edge case, I expect the user to set `createElement` and `recycleElement` within the same RAF.

PR #87, opened 2018-06-05
Start of ItemSource

domenic:
On top of #86; please look at only the last commit.

I only got this far before I ran out of time today :(. @valdrinkoshi, do you think you'd be able to help finish wiring this up? I wrote the docs and the ItemSource class itself, but haven't wired it through virtual-scroller-element and its underlying pieces yet.


item-source.js
```diff
@@ -0,0 +1,40 @@
+export const _getLength = new WeakMap();
+export const _item = new WeakMap();
+export const _key = new WeakMap();
```
domenic: I guess I should just use symbols since that's what the rest of the code is using. Let me switch it.

valdrinkoshi:
on it!

README.md
```diff
@@ -118,16 +108,48 @@ Fired when the scroller has finished rendering a new range of items, e.g. becaus
...
+
+There is also a factory method, `ItemSource.fromArray(array[, key])`, that makes this easier:
+
+```js
+const contactsSource = ItemSource.fromArray(contacts, c => c.id);
+```
+
+The `key` argument to `fromArray()` is called with an item, and should return a unique key for the object. If no `key` argument is given, then the item itself is used as the key.
+
+The main use of the `ItemSource` class is to be assigned to the `itemSource` property of a `<virtual-scroller>` element; as such, for now its only public API is a `length` property.
```
valdrinkoshi: `ItemSource` should also expose `item, key` properties so that `virtual-scroller` can access to them.

domenic: It can use the exported symbols, was my plan

valdrinkoshi: oh wow, I didn't even see they were exported! Awesome!

valdrinkoshi:
Done, we now consume ItemSource, demos updated as well.
If in the future we decide to expose `elementSource`, it will need to know about `itemSource`

item-source.js
```diff
@@ -0,0 +1,47 @@
...
+        return array[index];
+      },
+      key(index) {
+        return key ? key(array[index], index) : array[index];
+      }
+    });
+  }
+
+  get length() {
+    return this[_getLength]();
```
valdrinkoshi: Added `return`

item-source.js
```diff
@@ -0,0 +1,47 @@
...
+
+    return new this({
+      getLength() {
+        return array.length;
+      },
+      item(index) {
+        return array[index];
+      },
+      key(index) {
+        return key ? key(array[index], index) : array[index];
```
valdrinkoshi: Added `index` as second parameter, so that users wanting to map by index don't require to do expensive stuff like `ItemSource.fromArray(myItems, (item) => myItems.indexOf(item))`

README.md
```diff
@@ -118,16 +108,48 @@ Fired when the scroller has finished rendering a new range of items, e.g. becaus
 
 Also see [the example below](#performing-actions-as-the-scroller-scrolls-using-the-rangechange-event).
 
+### The `ItemSource` class
+
+The `ItemSource` class represents a way of translating indices into JavaScript values. You can create them like so:
+
+```js
+const source = new ItemSource({
```
valdrinkoshi: Is there a need to have a class? Implementation-wise, I feel this is not really needed. Also it looks more ergonomic to either accept an object or an array, e.g.
```js
// either object...
scroller.itemSource = {
  getLength() {},
  item(index) {},
  key(index) {}
};
// ...or array
scroller.itemSource = new Array(100);
```

virtual-scroller-element.js
```diff
@@ -1,18 +1,22 @@
+import {_item, _key, ItemSource} from './item-source.js';
```
valdrinkoshi: It feels weird to import these symbols here, all we need here is to check if the provided itemSource implements `item(), getLength(), key()` - maybe that can be an utility method?

README.md
```diff
@@ -48,43 +48,35 @@ Changing this property from its default will automatically reset `recycleElement
...
-Set this property to configure the virtual scroller with a function that will update the element with data at a given index.
+Set this property to configure the virtual scroller with a function that will update the element with data from a given item at the specified index.
 
 This property is invoked in these scenarios:
 
 * The user scrolls the scroller, changing which items' elements are visible. In this case, `updateElement` is called for all of the newly-visible elements.
-* The developer changes the `totalItems` property.
-* The developer calls `requestReset()`, which will call `updateElement` for all currently-visible elements. See [below](#data-manipulation-using-requestreset) for why this can be useful.
+* The developer changes the `itemSource` property.
+* The developer calls `itemsChanged()`, which will call `updateElement` for all currently-visible elements. See [below](#data-manipulation-using-itemschanged) for more on this.
```
kenchris: What if there is no recycling, should it then not update all elements?

README.md
kenchris: Any reason getLength is not a properly like Array.length?

README.md
```diff
@@ -118,16 +108,48 @@ Fired when the scroller has finished rendering a new range of items, e.g. becaus
...
+  key(index) { ... }
+});
+```
+
+For example, to create an `ItemSource` that gets its items from a `contacts` array, and uses `contact.id` as the key, you could do
+
+```js
+const contactsSource = new ItemSource({
+  item(index) { return contacts[index]; },
+  getLength() { return contacts.length; },
```
kenchris: Then `get length() { ... }`

README.md
valdrinkoshi: the elements should be updated in that case too, as currently visible data might have changed.

README.md
domenic: I think we need a class so that `list.itemSource` can return something useful and encapsulated and the same type each time.

virtual-scroller-element.js
domenic: The ItemSource doesn't expose those as public APIs. Those are capabilities that the ItemSource encapsulates, but we don't want to call the public methods of the ItemSource; in general classes in the web platform talk to each other through their internals, not their public methods.

README.md
domenic: That wouldn't work, because then we'd just invoke the getter once, in the constructor, and not have a reusable function we could call multiple times.

README.md
domenic: Also it allows us to have a simple factory, `ItemSource.fromArray(x, keyFunc)`, as well as potentially future other factories.

demo/sorting.html
```diff
@@ -39,35 +39,36 @@
...
       const newI = Math.max(0, Math.min(myContacts.length, i + offset));
       myContacts.splice(i, 1); // remove it
       myContacts.splice(newI, 0, contact); // add it to new position
-      virtualScroller.requestReset(); // notify virtual-scroller
+      virtualScroller.itemsChanged(); // notify virtual-scroller
     }
 
     window.toggleElementKey = function toggleElementKey(useGuid) {
-      virtualScroller.elementKey = useGuid ? (index) => myContacts[index].guid : null;
+      const key = useGuid ? (contact) => contact.guid : (_, index) => index;
```
domenic: Would contact work, or do we need contact.guid?

It feels like this demo is less interesting with the new architecture; to demo the bad version you actually have to explicitly make it have a bad key function. Not if that's worth fixing though.

smoke/infinite/index.html
```diff
@@ -20,14 +20,14 @@
...
-    const nodePool = [];
-    virtualScroller.createElement = () => nodePool.pop() || document.createElement('div');
-    virtualScroller.updateElement = (elem, i) => elem.textContent = i;
-    virtualScroller.recycleElement = (elem) => nodePool.push(elem);
     // Render as many items as possible --> infinite scrolling!
-    virtualScroller.totalItems = Number.MAX_SAFE_INTEGER;
+    virtualScroller.itemSource = new ItemSource({
+      getLength: () => Number.MAX_SAFE_INTEGER,
+      item: (index) => index,
+      key: (index) => index,
```
domenic: key shouldn't be necessary here

virtual-scroller-element.js
```diff
@@ -32,9 +36,12 @@ export class VirtualScrollerElement extends HTMLElement {
       }
       return childTemplate.cloneNode(true);
     };
-    this[_updateElement] = (element, index) => element.textContent = index;
+    this[_updateElement] = (element, item) => element.textContent =
+        item.toString();
     this[_recycleElement] = (element) => this[_nodePool].push(element);
-    this[_elementKey] = null;
+
+    this[_itemSource] = this[_rawItemSource] = null;
```
domenic: What is the point of _rawItemSource? Just to prevent re-rendering if you set the same object? I think it's OK in that case to re-render.

virtual-scroller-element.js
```diff
@@ -85,52 +92,70 @@ export class VirtualScrollerElement extends HTMLElement {
...
+        itemSource;
+    this[_render]();
   }
 
   get createElement() {
     return this[_createElement];
   }
   set createElement(fn) {
+    // No change.
+    if (fn === this[_createElement]) {
```
domenic: In general I'm not sure of the value of these kind of guards.

virtual-scroller-element.js
valdrinkoshi: Uhm, I see. I like it!
So if I wanted to make those methods public, I'd do something like this, right?
```js
import {ItemSource, _key, _item} from 'item-source.js';
class ContactSource extends ItemSource {
  get key() {
    return this[_key];
  }
  get item() {
    return this[_item];
  }
}
```

virtual-scroller-element.js
valdrinkoshi: Yes, it's avoiding rerendering and invalidation of the [internal key/Element map](https://github.com/Polymer/lit-element#minimal-example).
ItemSource determines the `elementKey` function, which when gets changed causes invalidation of the internal map.

This would be a problem in patterns like LitElement where we always set properties/attributes when rendering/updating the dom - see https://github.com/Polymer/lit-element#minimal-example
As an example, imagine I'm rendering virtual-scroller and setting the contacts **array**
```js
_render({ contacts }) { // typeof contacts === 'Array' 
  return html`<virtual-scroller itemSource="${contacts}"></virtual-list>`;
}
```

virtual-scroller-element.js
valdrinkoshi: Avoid rendering. In particular `createElement` changes cause the internal key/Element map to be invalidated.

item-source.js
```diff
@@ -0,0 +1,47 @@
...
+
+export class ItemSource {
+  constructor({getLength, item, key}) {
+    if (typeof getLength !== 'function') {
+      throw new TypeError('getLength option must be a function');
+    }
+    if (typeof item !== 'function') {
+      throw new TypeError('item option must be a function');
+    }
+    if (typeof key !== 'function') {
```
valdrinkoshi: key is mandatory here - should it be?

smoke/infinite/index.html
valdrinkoshi: it's currently mandatory - should it?

virtual-scroller-element.js
domenic: Well, I don't think we'd expose the _key and _item to web developers. This is more just me being conservative until we know what would be useful for web developers: if we have use cases where item() or key() are useful, we should just expose them directly on ItemSource. (Also, I'd probably tweak things to use a proxy so that instead of source.item(3) you could just do source[3].)

smoke/infinite/index.html
domenic: Oh, right, it is. I confused myself because I changed this a couple times locally before uploading the current version.

I think it's OK for it to be mandatory since the constructor is kind of the "low level" raw API. At least for now.

valdrinkoshi:
LGTM too - small question regarding `key()` if it should be mandatory or not

virtual-scroller-element.js
domenic: Oh, wow, so how does LitElement handle things like `imgElement.src`, where setting it causes a whole new fetch?

virtual-scroller-element.js
domenic: Sure. So don't set createElement if you don't want things to be re-rendered, is my argument. (But, if there are popular frameworks which make that hard, then we need to think about it. Let's discuss in the above thread about img.src etc.)

virtual-scroller-element.js
valdrinkoshi: it does check for identity internally before actually updating the property/attribute.
In our case, the getter always returns an ItemSource object, whereas the setter receives an Array.

virtual-scroller-element.js
valdrinkoshi: sg, let me remove these

virtual-scroller-element.js
domenic: Ah, interesting point. Happy to leave in the check for now and think harder later. Maybe we should reconsider that automatic conversion? Or maybe this is just a special case though and keeping the check is OK.

valdrinkoshi: I think it would be better to return the rawItemSource in the getter and use the actual ItemSource internally. Like that, the user will get what he's set. We'll still need to do the identity check to know if to update the internal ItemSource tho

valdrinkoshi: I'll open an issue on this, so we can go ahead and merge here 👌

valdrinkoshi: Filed https://github.com/valdrinkoshi/virtual-scroller/issues/93 - feel free to merge this!

PR #96, opened 2018-06-12
Add demo: list with dismissable items

kenchris:
Add an example of how you can create a list with dismissable/deleteable items.

### DIFF
--- a/demo/dismissable/dismissable-item.js
+++ b/demo/dismissable/dismissable-item.js
@@ -32,10 +32,6 @@ class VelocityTracker {
 class DismissableItem extends HTMLElement {
   constructor() {
     super();
-    this.attachShadow({ mode: 'open' });
-    this.shadowRoot.innerHTML = `
-      <slot></slot>
-    `;
 
     this.position = 0;
     this.itemIndex = 0;
@@ -81,6 +77,7 @@ class DismissableItem extends HTMLElement {
   }
 
   setPosition(position) {
+    this.position = position;
     this.width = this.offsetWidth;
     this.style.opacity = (this.width - Math.abs(position)) / this.width;
     const currentY = this.style.transform.split(',')[1];
@@ -177,6 +174,8 @@ class DismissableItem extends HTMLElement {
       const deltaX = change.clientX - this.startX;
       const deltaY = change.clientY - this.startY;
 
+      // Shaky or sloppy fingers are common using scroll, so
+      // we ignore mini pans of 5 dips like Android.
       if (deltaX ** 2 + deltaY ** 2 < kTouchSlopValue ** 2) {
         e.preventDefault();
         return;
@@ -197,8 +196,7 @@ class DismissableItem extends HTMLElement {
 
       e.preventDefault();
       const deltaX = change.clientX - this.startX;
-      this.position = this.startPosition + deltaX;
-      this.setPosition(this.position);
+      this.setPosition(this.startPosition + deltaX);
     }
   }
 
--- a/demo/dismissable/dismissable-item.js
+++ b/demo/dismissable/dismissable-item.js
@@ -20,7 +20,7 @@ class VelocityTracker {
 
     const oldestTouchMove = this._recentTouchMoves[0];
 
-    const deltaX = e.changedTouches[0].clientX - oldestTouchMove.changedTouches[0].clientX;
+    const deltaX = e.clientX - oldestTouchMove.clientX;
     const deltaT = e.timeStamp - oldestTouchMove.timeStamp;
 
     return {
@@ -33,27 +33,37 @@ class DismissableItem extends HTMLElement {
   constructor() {
     super();
     this.attachShadow({ mode: 'open' });
-    this.shadowRoot.innerHTML = '<slot></slot>';
+    this.shadowRoot.innerHTML = `
+      <style>
+      :host {
+        touch-action: none;
+      }
+      </style>
+      <slot></slot>
+    `;
 
     this.position = 0;
     this.itemIndex = 0;
     this.width = 0;
     this.state = 'initial';
-    this.addEventListener('touchstart', this);
-    this.addEventListener('touchmove', this);
-    this.addEventListener('touchend', this);
+    this.addEventListener('pointerdown', this);
+    this.addEventListener('pointermove', this);
+    this.addEventListener('pointerup', this);
   }
 
   handleEvent(event) {
     switch(event.type) {
-      case 'touchstart':
-        this._onTouchStart(event);
+      case 'pointerdown':
+        this._onPointerDown(event);
         break;
-      case 'touchmove':
-        this._onTouchMove(event);
+      case 'pointermove':
+        if (!event.pressure) {
+          break;
+        }
+        this._onPointerPan(event);
         break;
-      case 'touchend':
-        this._onTouchEnd(event);
+      case 'pointerup':
+        this._onPointerUp(event);
         break;
     }
   }
@@ -143,17 +153,18 @@ class DismissableItem extends HTMLElement {
     }
   }
 
-  _onTouchStart(e) {
+  _onPointerDown(e) {
     this.state = 'initial';
-    this.startX = e.changedTouches[0].clientX;
-    this.startY = e.changedTouches[0].clientY;
+    this.setPointerCapture(e.pointerId);
+    this.startX = e.clientX;
+    this.startY = e.clientY;
     this.startPosition = 0;
   }
 
-  _onTouchMove(e) {
+  _onPointerPan(e) {
     if (this.state == 'initial') {
-      const deltaX = e.changedTouches[0].clientX - this.startX;
-      const deltaY = e.changedTouches[0].clientY - this.startY;
+      const deltaX = e.clientX - this.startX;
+      const deltaY = e.clientY - this.startY;
 
       if (deltaX ** 2 + deltaY ** 2 < kTouchSlopValue ** 2) {
         e.preventDefault();
@@ -174,13 +185,14 @@ class DismissableItem extends HTMLElement {
       this._tracker.update(e);
 
       e.preventDefault();
-      const deltaX = e.changedTouches[0].clientX - this.startX;
+      const deltaX = e.clientX - this.startX;
       this.position = this.startPosition + deltaX;
       this.setPosition(this.position);
     }
   }
 
-  _onTouchEnd(e) {
+  _onPointerUp(e) {
+    this.releasePointerCapture(e.pointerId);
     if (this.state == 'dragging') {
       const velocity = this._tracker.update(e).velocityX;
       if (Math.abs(velocity) > kMinFlingVelocityValue) {
--- a/demo/dismissable/dismissable-item.js
+++ b/demo/dismissable/dismissable-item.js
@@ -181,16 +181,15 @@ class DismissableItem extends HTMLElement {
   }
 
   _onTouchEnd(e) {
-    const velocity = this._tracker.update(e).velocityX;
-    this._tracker = null;
-
     if (this.state == 'dragging') {
+      const velocity = this._tracker.update(e).velocityX;
       if (Math.abs(velocity) > kMinFlingVelocityValue) {
         this.fling(velocity);
         return;
       }
       this._settleToClosestPosition();
     }
+    this._tracker = null;
   }
 }
 
--- a/demo/dismissable/dismissable-item.js
+++ b/demo/dismissable/dismissable-item.js
@@ -1,0 +1,233 @@
+const kMinFlingVelocityValue = 0.4;
+const kTouchSlopValue = 5;
+
+class VelocityTracker {
+  constructor(target) {
+    this.target = target;
+
+    this._recentTouchMoves = [];
+    this.velocityX = 0;
+    this._timeWindow = 50;
+
+    this._onTouchStart = this._onTouchStart.bind(this);
+    this._onTouchMove = this._onTouchMove.bind(this);
+    this._onTouchEnd = this._onTouchEnd.bind(this);
+  }
+
+  start() {
+    this.target.addEventListener('touchstart', this._onTouchStart);
+    this.target.addEventListener('touchmove', this._onTouchMove);
+    this.target.addEventListener('touchend', this._onTouchEnd);
+  }
+
+  stop() {
+    this._recentTouchMoves = [];
+    this.velocityX = 0;
+    this.target.removeEventListener('touchstart', this._onTouchStart);
+    this.target.removeEventListener('touchmove', this._onTouchMove);
+    this.target.removeEventListener('touchend', this._onTouchEnd);
+  }
+
+  _pruneHistory(timeStamp) {
+    for (let i = 0; i < this._recentTouchMoves.length; ++i) {
+      if (this._recentTouchMoves[i].timeStamp > timeStamp - this._timeWindow) {
+        this._recentTouchMoves = this._recentTouchMoves.slice(i);
+        return;
+      }
+    }
+    this._recentTouchMoves = [];
+  }
+
+  _update(e) {
+    this._pruneHistory(e.timeStamp);
+    this._recentTouchMoves.push(e);
+
+    const oldestTouchMove = this._recentTouchMoves[0];
+
+    const deltaX = e.changedTouches[0].clientX - oldestTouchMove.changedTouches[0].clientX;
+    const deltaT = e.timeStamp - oldestTouchMove.timeStamp;
+
+    if (deltaT > 0) {
+      this.velocityX = deltaX / deltaT;
+    } else {
+      this.velocityX = 0;
+    }
+  }
+
+  _onTouchStart(e) {
+    this._recentTouchMoves.push(e);
+    this.velocityX = 0;
+  }
+
+  _onTouchMove(e) {
+    this._update(e);
+  }
+
+  _onTouchEnd(e) {
+    this._update(e);
+    this._recentTouchMoves = [];
+  }
+}
+
+class DismissableItem extends HTMLElement {
+  constructor() {
+    super();
+    this.attachShadow({ mode: 'open' });
+    this.shadowRoot.innerHTML = `
+      <div>
+        <slot></slot>
+      </div>
+    `;
+
+    this.tracker = new VelocityTracker(this);
+
+    this.position = 0;
+    this.itemIndex = 0;
+    this.width = 0;
+
+    this.state = "initial";
+
+    this.addEventListener('touchstart', this._onTouchStart.bind(this));
+    this.addEventListener('touchmove', this._onTouchMove.bind(this));
+    this.addEventListener('touchend', this._onTouchEnd.bind(this));
+  }
+
+  setPosition(position) {
+    this.width = this.offsetWidth;
+    this.style.opacity = (this.width - Math.abs(position)) / this.width;
+    const currentY = this.style.transform.split(',')[1];
+    this.style.transform = `translate(${position}px, ${currentY}`;
+  }
+
+  _dismiss() {
+    this.style.opacity = 0;
+    const height = getComputedStyle(this).height;
+
+    const collapseAnim = this.animate([
+      { height }, { height: '0px'}
+    ], { 
+      duration: 100,
+      iterations: 1
+    });
+
+    collapseAnim.onfinish = () => {
+      this.setPosition(0);
+      const event = new CustomEvent('remove', { index: this.itemIndex });
+      this.dispatchEvent(event);
+    }
+  }
+
+  settle(targetPosition) {
+    this.state = "initial";
+
+    const basePosition = this.position;
+    if (targetPosition === basePosition) {
+      return;
+    }
+
+    this.position = targetPosition;
+    const baseOpacity = this.style.opacity;
+
+    const currentY = this.style.transform.split(',')[1];
+
+    const isDismiss = targetPosition !== 0;
+
+    const animation = this.animate([
+      { transform: `translate(${basePosition}px,${currentY}`, opacity: baseOpacity },
+      { transform: `translate(${targetPosition}px,${currentY}`, opacity: isDismiss ? 0 : 1}
+    ], { 
+      duration: Math.abs(targetPosition - basePosition) * 0.5,
+      iterations: 1
+    });
+
+    animation.onfinish = () => {
+      if (isDismiss) {
+        this._dismiss();
+      } else {
+        this.setPosition(0);
+      }
+    }
+  }
+
+  fling(velocityX) {
+    this.state = "initial";
+    const basePosition = this.position;
+    const targetPosition = velocityX < 0 ? -this.width : this.width;
+
+    const baseOpacity = this.style.opacity;
+
+    const currentY = this.style.transform.split(',')[1];
+
+    const animation = this.animate([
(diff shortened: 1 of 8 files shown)


kenchris:
Video of how it works: https://www.youtube.com/watch?v=NLkt1t2gWNI

demo/dismissable/dismissable-item.js
```diff
@@ -0,0 +1,233 @@
...
+    this.state = "initial";
+    const basePosition = this.position;
+    const targetPosition = velocityX < 0 ? -this.width : this.width;
+
+    const baseOpacity = this.style.opacity;
+
+    const currentY = this.style.transform.split(',')[1];
+
+    const animation = this.animate([
+      { transform: `translate(${basePosition}px,${currentY}`, opacity: baseOpacity },
```
valdrinkoshi: hopefully we can get away without defining the "from" keyframe here as well

demo/dismissable/dismissable-item.js
```diff
@@ -0,0 +1,233 @@
...
+    const height = getComputedStyle(this).height;
+
+    const collapseAnim = this.animate([
+      { height }, { height: '0px'}
+    ], { 
+      duration: 100,
+      iterations: 1
+    });
+
+    collapseAnim.onfinish = () => {
```
valdrinkoshi: nit: make this method a new property `_fireRemove()`

kenchris:
Lots of great comments, thank you! I will have a look at them all tomorrow morning.

> would be nice to get some details/comments on how these parts work together.

Yes, I am still trying to find out the best and most seamless way to make it all work together and simplify it, but I thought it was better to upload my first fully working implementation that I was reasonable satisfied with :-) 

> How's the behavior on the actual device?

I tested an earlier version yesterday night and it worked fine on my Pixel 2. I tried to match the feeling from other native Android apps.

> Do you think we can enable these gestures also on desktop through mouseup/down/move? Maybe not worth it?

It shouldn't be that hard to implement, and I could potentially use pointer events. Currently it is not a very common user interaction model on desktop, but many desktop apps need to support both keyboard and touch (Windows 10, Pixelbook etc) so that could of course change over time, as users become used to the pattern.

demo/dismissable/dismissable-item.js
```diff
@@ -0,0 +1,233 @@
...
+    const currentY = this.style.transform.split(',')[1];
+    this.style.transform = `translate(${position}px, ${currentY}`;
+  }
+
+  _dismiss() {
+    this.style.opacity = 0;
+    const height = getComputedStyle(this).height;
+
+    const collapseAnim = this.animate([
+      { height }, { height: '0px'}
```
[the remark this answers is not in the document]
kenchris: I will have to try again, but I got some error messages about non matching keyframes. I will need to read the spec so see why that is.

demo/dismissable/index.html
```diff
@@ -0,0 +1,72 @@
...
+
+  <script type="module">
+    import { ItemSource } from '../../virtual-scroller-element.js';
+    import './../contacts/contact-element.js';
+    import './dismissable-item.js';
+
+    const virtualScroller = document.querySelector('virtual-scroller');
+
+    const createElementFn = virtualScroller.createElement.bind(virtualScroller);
+    virtualScroller.createElement = (item, itemIndex) => {
```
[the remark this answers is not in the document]
kenchris: Are all current event listeners removed from an element when it is pushed to the element pool? Or that happens because it gets detached from the DOM?

I believe that adding the same function for the same event, should not duplicate the listener, so doing this should be fine.

demo/dismissable/dismissable-item.js
```diff
@@ -0,0 +1,233 @@
...
+
+  stop() {
+    this._recentTouchMoves = [];
+    this.velocityX = 0;
+    this.target.removeEventListener('touchstart', this._onTouchStart);
+    this.target.removeEventListener('touchmove', this._onTouchMove);
+    this.target.removeEventListener('touchend', this._onTouchEnd);
+  }
+
+  _pruneHistory(timeStamp) {
```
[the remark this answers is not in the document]
kenchris: Good idea, let me do that

demo/dismissable/dismissable-item.js
```diff
@@ -0,0 +1,233 @@
...
+
+    this.position = 0;
+    this.itemIndex = 0;
+    this.width = 0;
+
+    this.state = "initial";
+
+    this.addEventListener('touchstart', this._onTouchStart.bind(this));
+    this.addEventListener('touchmove', this._onTouchMove.bind(this));
+    this.addEventListener('touchend', this._onTouchEnd.bind(this));
```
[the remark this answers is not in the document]
kenchris: Will do.

demo/dismissable/index.html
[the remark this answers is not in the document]
valdrinkoshi: They're not removed, but as you noted we'd be adding the same function for the same event and get deduplication from the browser.

Alternatively, we could make `remove` event `bubble: true`, and have a single listener on virtualScroller - that should look even cleaner.
```js
// `remove` event bubbles
virtualScroller.addEventListener('remove', (event) => {
  const dismissable = event.target;
  myContacts.splice(dismissable.itemIndex, 1);
  virtualScroller.itemsChanged();
});
virtualScroller.updateElement = (element, contact, index) => {
  element.children[0].contact = contact;
  element.itemIndex = index;
};

File: demo/dismissable/dismissable-item.js (hunk 3 above)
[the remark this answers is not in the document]
Comment (kenchris) [reply demo/dismissable/dismissable-item.js]: Unfortunately no :-(

```
Uncaught DOMException: Failed to execute 'animate' on 'Element': Partial keyframes are not supported.
    at HTMLElement._dismiss (http://linuxhost:8000/demo/dismissable/dismissable-item.js:76:31)
    at Animation.animation.onfinish (http://linuxhost:8000/demo/dismissable/dismissable-item.js:121:14)
```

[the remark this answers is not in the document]
Comment (kenchris) [reply demo/dismissable/dismissable-item.js]: https://github.com/web-animations/web-animations-js/issues/14
https://github.com/web-animations/web-animations-next/issues/325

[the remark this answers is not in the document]
Comment (kenchris) [reply demo/dismissable/dismissable-item.js]: Oh, that seems to be for the polyfill which I am not using

**Comment by kenchris** [comment]
OK, fixed the review comments and it looks much nicer. I don't have an iPhone so couldn't test on iOS, but Web Animations is behind a flag there so either a polyfill is needed or I need to just animate using CSS.

I did try that, but for some reason it doesn't animate to the end and I never got the `transitionend` event. It is also quite annoying as I only want animations in one direction (collapse) so I need to add and remove CSS classes to remove the transitions - so in that sense Web Animations is much nicer.

File: demo/dismissable/dismissable-item.js (hunk 3 above)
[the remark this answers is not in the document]
Comment (kenchris) [reply demo/dismissable/dismissable-item.js]: According to spec https://drafts.csswg.org/web-animations/ something like

`this.animate({ transform: 'scale(0)', opacity: 0 }, 300);`

should work, but fails in Chrome

**Comment by kenchris** [comment]
I didn't include my changes to the deployment script, because it doesn't work by default for me - I am on Linux so it might be due to `sed` differences between Linux and Mac

My changes are:
```
+mv node_modules/web-animations-js components/
 # replace node_modules/ with components/
-find node_modules/$repo -name '*.js' -exec sed -i '' 's/node_modules\//components\//g' {} \;
+find node_modules/$repo -name '*.js' -exec sed -i 's/node_modules\//components\//g' {} \;
+find -name '*.html' -exec sed -i 's/node_modules\//components\//g' {} \;
```

**Comment by kenchris** [comment]
So moving to pointer events, made the list not scroll on mobile as it relies on regular touch events. Maybe we could consider moving to pointer events for the list and that would solve this issue.

The reason is that in order to use *pointer events* for *touch*, you must set `touch-action: none` to block
*regular touch events*, and *regular touch events* **cancels** *touch based pointer events* :-(

My last patch uses only pointer events for mouse.

**Comment by domenic** [comment]
> So moving to pointer events, made the list not scroll on mobile as it relies on regular touch events. Maybe we could consider moving to pointer events for the list and that would solve this issue.

This is interesting, and might be worth filing as a separate issue. I hadn't realized we were using touch events...

(But to be clear, we shouldn't necessarily block on figuring this out before landing this cool demo.)

File: demo/dismissable/dismissable-item.js [hunk 7]
```diff
@@ -0,0 +1,218 @@
...
+    return {
+      velocityX: (deltaT > 0) ? deltaX / deltaT : 0
+    };
+  }
+}
+
+class DismissableItem extends HTMLElement {
+  constructor() {
+    super();
+    this.attachShadow({ mode: 'open' });
```
valdrinkoshi: seems like we don't really need a shadowRoot for this element

kenchris:
Actually, the current code works fine.

It wasn't working when I just used pointer events, because touch events cancel pointer events (ie. `pointercancel` gets called after like 5-10 `pointermove` events). The only way to avoid that, as far as I researched, is to set `touch-action: none`, but that means that no touch events are forwarded to virtual-scroller, and only pointer move events, which virtual-scroller doesn't seem to handle.

demo/dismissable/dismissable-item.js
```diff
@@ -0,0 +1,218 @@
...
+    this.startPosition = 0;
+  }
+
+  _onPointerPan(e, change) {
+    if (this.state == 'initial') {
+      const deltaX = change.clientX - this.startX;
+      const deltaY = change.clientY - this.startY;
+
+      if (deltaX ** 2 + deltaY ** 2 < kTouchSlopValue ** 2) {
+        e.preventDefault();
```
valdrinkoshi: If I understood this correctly, we prevent the default if there was not enough movement. Why is this needed?

I would rather check if we are scrolling vertically or horizontally, and prevent default if really needed thereafter, something like this maybe:
```js
_onTouchStart(event) {
   // ...
   this.isHorizontalSwipe = null;
}
_onTouchMove(event) {
  if (this.isHorizontalSwipe === null) {
    const deltaX = Math.abs(change.clientX - this.startX);
    const deltaY = Math.abs(change.clientY - this.startY);
    // not enough movement to determine direction!
    if (deltaX < kTouchSlopValue  && deltaY < kTouchSlopValue) {
      return;
    }
    this.isHorizontalSwipe = deltaX >= deltaY;
  }
  // scrolling vertically, ignore this and following events
  if (this.isHorizontalSwipe === false) {
    return;
  }
  // scrolling horizontally!
```

demo/dismissable/dismissable-item.js
```diff
@@ -0,0 +1,218 @@
...
+      this.width = this.offsetWidth;
+    }
+
+    if (this.state == 'dragging') {
+      this._tracker = this._tracker || new VelocityTracker();
+      this._tracker.update(e);
+
+      e.preventDefault();
+      const deltaX = change.clientX - this.startX;
+      this.position = this.startPosition + deltaX;
```
valdrinkoshi: nit: this should be done in `setPosition()` rather than here

kenchris:
From: https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action

> An application using Pointer_events will receive a pointercancel event when the browser starts handling a touch gesture. 

This seems to be what I am running into.

demo/dismissable/dismissable-item.js
kenchris: Sloppy or shaky fingers are quite common

kenchris: This is also what is done on Android and we did the same when I worked at Nokia :-)

You can see some examples here:
https://developer.android.com/training/gestures/viewgroup

demo/dismissable/dismissable-item.js
kenchris: You mean the last line? like `setPosition(this.startPosition + deltaX)` ?

demo/dismissable/dismissable-item.js
kenchris: Right, I was thinking the same earlier. Done

demo/dismissable/dismissable-item.js
kenchris: Fixed

kenchris:
@valdrinkoshi Did you want to fix the deploy script yourself? I need to change node_modules to components in the .html file and the first '' in the `sed` command makes the `sed` fail for me on Chrome OS (Termina VM Container)

valdrinkoshi:
Uhm, I don't think we need the polyfill for Web Animations, as the demo won't work in iOs anyways, as VirtualScroller depends on ResizeObserver. I opened this issue to track what features we're using https://github.com/valdrinkoshi/virtual-scroller/issues/100

kenchris:
OK, I can remove it...

demo/dismissable/dismissable-item.js
```diff
@@ -0,0 +1,216 @@
...
+
+  _onPointerPan(e, change) {
+    if (this.state == 'initial') {
+      const deltaX = change.clientX - this.startX;
+      const deltaY = change.clientY - this.startY;
+
+      // Shaky or sloppy fingers are common using scroll, so
+      // we ignore mini pans of 5 dips like Android.
+      if (deltaX ** 2 + deltaY ** 2 < kTouchSlopValue ** 2) {
+        e.preventDefault();
```
valdrinkoshi: not clear why we preventDefault here 🤔 
This one or the one below seem to cause browser intervention when scrolling vertically more than horizontally, e.g. 
scroll-error

This prompted the earlier suggestion to check if we're scrolling horizontally or not, and avoid updating position/opacity if the user is scrolling vertically.

demo/dismissable/dismissable-item.js
```diff
@@ -0,0 +1,216 @@
...
+        return;
+      }
+
+      if (Math.abs(deltaY) > Math.abs(deltaX)) {
+        this._settleToClosestPosition();
+        return;
+      }
+
+      this.state = 'dragging';
+      this.width = this.offsetWidth;
```
valdrinkoshi: why do we remeasure the width here?

demo/dismissable/dismissable-item.js
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    if (this.state == 'dragging') {
+      this._tracker = this._tracker || new VelocityTracker();
+      this._tracker.update(e);
+
+      e.preventDefault();
```
valdrinkoshi: not clear why this is needed here

demo/dismissable/dismissable-item.js
kenchris: The second preventDefault avoid the following:

https://www.youtube.com/watch?v=f5KsaMW4RW4

Maybe we can fix that a better way

valdrinkoshi: oh interesting...I think that _should_ be fixable by setting
```css
virtual-scroller {
  overflow-x: hidden;
}

File: demo/dismissable/dismissable-item.js (hunk 11 above)
Comment (kenchris) [reply demo/dismissable/dismissable-item.js]: Probably a leftover from refactoring, removed

File: demo/dismissable/dismissable-item.js (hunk 10 above)
Comment (kenchris) [reply demo/dismissable/dismissable-item.js]: The first perventDefaults can be removed.

I did run into this other bug though, which I have to find out why happens:

https://drive.google.com/file/d/1irkYq-XyI7C-HLedtXlms12D-cJfo1AX/view

Comment (valdrinkoshi) [reply demo/dismissable/dismissable-item.js]: I assume that happens because, towards the end of that gesture, `deltaX` becomes finally bigger than `deltaY`. That gesture starts as a vertical scroll, and ends as a horizontal one.
I'm fine keeping it like that for now, but if you don't like it we should detect somehow if the gesture starts as a horizontal pan or not

Comment (kenchris) [reply demo/dismissable/dismissable-item.js]: yes, `overflow-x: hidden` will work, though I wish it could be done on the `dismissable-item` instead for better encapsulation

Comment (kenchris) [reply demo/dismissable/dismissable-item.js]: Actually if the list starts scrolling, that should cancel the dismissal handling. 

If I start a drag horizontally, then I cannot scroll until I lift my finger/mouse.. even if I pan my finger all the way down vertically and then to the right. That works like Android - just try panning away one of the notifications on Android - press, move a bit left or right, then while finger down, move all the way down and then you can still dismiss by moving left and right.

I don't want to break that behavior. So basically just knowing when the list is being scrolled and then dismiss would be fine!

**Comment by kenchris** [comment]
Putting here as GitHub collapses old comments:

Actually if the list starts scrolling, that should cancel the dismissal handling.

If I start a drag horizontally, then I cannot scroll until I lift my finger/mouse.. even if I pan my finger all the way down vertically and then to the right. That works like Android - just try panning away one of the notifications on Android - press, move a bit left or right, then while finger down, move all the way down and then you can still dismiss by moving left and right.

I don't want to break that behavior. So basically just knowing when the list is being scrolled and then dismiss would be fine!

File: demo/dismissable/dismissable-item.js [hunk 13]
```diff
@@ -0,0 +1,219 @@
...
+      const deltaY = change.clientY - this.startY;
+
+      // Shaky or sloppy fingers are common using scroll, so
+      // we ignore mini pans of 5 dips like Android.
+      if (deltaX ** 2 + deltaY ** 2 < kTouchSlopValue ** 2) {
+        return;
+      }
+
+      if (Math.abs(deltaY) > Math.abs(deltaX)) {
+        this._settleToClosestPosition();
```
valdrinkoshi: oh! @kenchris I think all we need to do to cancel the dismissal is to set a new state in here, since in this case we captured the fact that there is more vertical scroll than horizontal.
Adding a line here should do the trick:
```js
this._settleToClosestPosition();
this.state = 'cancel';
```
The rest of the code already handles the resetting of the state, and will do position/opacity if state == 'start' or state == 'dragging' only 👌

valdrinkoshi:
Agreed on the design, I think adding the state = 'cancel' in the right spot should fix the last issue you mentioned \o/

demo/dismissable/dismissable-item.js
kenchris: This still doesn't solve that you can sometimes start a horizontal pan after starting to scroll:

https://drive.google.com/file/d/17wW9SmR9WlxwN9uqfvFf9xAvIibr_DZe/view

Also as you see the `overflow-x: hidden` trick doesn't seem to be enough

kenchris: Actually you might be right, it seems that the change maybe wasn't reloaded properly. Now I cannot seem to trigger it anymore \o/

kenchris: No I can still trigger it, by quickly moving down or up in a 90 degree angle - that still start list scrolling and panning

kenchris: If I readd that e.preventDefault() inside the dragging/panning code, then no scroll can happen at the same time that that solves this. Can I re-add that?

valdrinkoshi: Argh :x 
This needs more investigation to be perfected. 
I'll have some time to look more into it later, I was thinking to see if keeping only touch* events solves the problem.

kenchris: @valdrinkoshi great. I will upload the preventDefault() fix as that is all working. If you end up fixing it in any other way that is great with me :-)

kenchris:
Interesting observation. You can dismiss multiple contacts using multiple fingers. It actually seem to work, but I see some jank.

I checked Android and apps only seem to allow one dismissable item at the time. In Google Inbox... multiple touch points means that the dismissal gets ignored outright.

For notifications, though, one of the fingers will be used, so you can dismiss one item - Google Tasks and GMail works the same way. - that seems to be the default.

valdrinkoshi:
I've played with this a bit, sent a PR on your repo https://github.com/kenchris/virtual-scroller/pull/1
I think some of the problems might have been related to not resetting `_tracker` when gesture finished. anyways, on my local tests i don't face the issues you mentioned

kenchris:
I can still trigger the problem though... just move more in x direction than y: 

https://drive.google.com/file/d/1UVg8LCaqQ93vHmGm57PCeEq72KgEG2Zu/view

The DevTools shows that it is running the newest code

kenchris:
Yay, thanks for merging \o/

valdrinkoshi:
Interesting...i see in your console that `content-script.js` is causing a TypeError right before you see that problem...that error in particular seems to be doing something with ShadowRoots - should be some extension you have active. Can you try it on a clean chrome profile, or by disabling extensions?

kenchris:
Yes, that is when I use the recording extension... that is also the one that breaks the `overflow-x:hidden`.

I can easily reproduce with that extension disabled - but then I just cannot record on Chrome OS :-)

kenchris:
When I am in dragging/panning mode, I really don't want anyone else acting on the same events, to do scrolling and whatnot. I tried `event.stopPropagation()` but only `event.preventDefault()` is working for that.

kenchris:
I guess that it is not the worst in the world that you can scroll while panning, if it wasn't for the fact that the draggable state switches between 0 and the drag position and the virtual list resets the transform during scroll.

valdrinkoshi:
oh...what about updating the transform of the distributed `<content-element>` instead of the `<dismissable-item>`? That way we could scope all the things within `<dismissable-item>`, also the `overflow-x: hidden`

kenchris:
hmm, that is one option. But then I could only have one child or would need to apply transform to them all, or is there a smarter way. I could add the `overflow-x: hidden` inside the shadow root.

kenchris:
I really don't know how this is happening.. for a second I though it was the `deltaX ** 2 + deltaY ** 2 < kTouchSlopValue ** 2` code that allow a scroll to start, but even removing that and I can still trigger it

valdrinkoshi:
Uhm, if we want to support multiple children, then the shadowRoot of `<dismissable-item>` could look like this:
```html
<style>
  #contentWrapper {
    overflow-x: hidden;
   }
</style>
<div id="contentWrapper">
  <slot></slot>
</div>
```
And you could then translate `#contentWrapper`

valdrinkoshi:
Oh forgot to answer to your question
> May I ask why you don't like it? It is only being called in the panning code.

In general, I'm afraid of preventDefault() on scroll-related events, as it might prevent the browser from performing smooth scrolls. e.g. chrome made certain event listeners passive by default to improve scroll performance on mobile https://developers.google.com/web/tools/lighthouse/audits/passive-event-listeners

In this case, I'd like to understand what is causing that scroll that we're trying to prevent, and see if we can get around it with css. If we can do that, the next step would be to just listen for those events with `{passive: true}` ~o~

kenchris:
OK that works nicer: https://github.com/valdrinkoshi/virtual-scroller/pull/107

kenchris:
Yes, I know about the passive listeners and the interventions. cc @rbyers 

Here in this case we don't want a scroll because we are in panning/draggable mode - so yes, we are preventing a scroll here. Also, this preventDefault() is only called while panning/dragging.

I assume that wouldn't affect any future scroll, for instance started with a touch flick (touchstart, touchmove, touchend without any call to preventDetault(). Maybe Rick can enlighten us here :-)

valdrinkoshi:
Thanks for the hard work @kenchris - this little demo is becoming a great custom element for swiping stuff away \o/

kenchris:
Indeed :-) I really want our nice web platform to have these nice primitives available, working well and easy to use and re-use.

I still hope we can find and fix this last bug in a nice manner though, but I first need to figure out what is going on.

Issue #64, opened 2018-05-09
scroll position is not kept on resize


valdrinkoshi:
To reproduce
- go to https://valdrinkoshi.github.io/virtual-list/demo/contacts/
- scroll down a couple items
- resize the window
- note that the scroll position gets offset by items changing their size

![bug-scroll-position](https://user-images.githubusercontent.com/6173664/39830388-26778f74-5376-11e8-8633-8b7d581893aa.gif)

valdrinkoshi:
note to self: there's probably a bug on how we keep track of the first item offset - checkout https://developers.google.com/web/updates/2016/07/infinite-scroller#scroll_anchoring

valdrinkoshi:
Layout already ensures that the scroll position is kept, and uses the first **rendered** item as an anchor. When _overhang > 0, items get positioned outside the viewport, so that they can be rendered before appearing into the screen when scrolling.

The example above shows a different requirement, where the user wants to use the first **visible** item as an anchor. This could be achieved by setting _overhang to 0, but the scrolling performance might suffer from it.

I can imagine other use cases where the user wants to use another item as an anchor, e.g. in a chat app, use the last item as the anchor when user scrolls to the bottom. Or in an horizontal carousel, use the center item as the anchor. In general, user requirement is that other items resize or window resize push the other items around the anchor item.

Would it make sense to rather expose a property for this? I think all we need is some sort of `anchorOffset: number`, which can be set as:
- `anchorOffset = overhang` => uses the first visible item as an anchor
- `anchorOffset = overhang + viewportSize / 2` => center item as an anchor
- `anchorOffset = overhang + viewportSize` => last item as an anchor

@domenic @graynorton wdyt?

graynorton:
A couple of questions:

1) Do you envision this property as existing only on the layout, or would it be facaded up to `virtual-scroller` itself?

2) While the numeric value offers complete flexibility, it seems like it might be a bit fiddly and/or error-prone. In addition to numeric values, what if we were to accept a small number of string values for common cases like `start`, `middle` and `end`? If the user chose one of these, the implementation could do the math.

domenic:
I tend to agree that a common-case enum of those three options would be best. I'm not sure the full flexibility of anchorOffset is needed.

@tabatkins do I recall there being some CSS proposal that controlled this type of thing? I.e. that allowed the scrollbar to "stick" to the bottom when resizing something like a chat window?

Also /cc @flackr for his thoughts.

[...]

PR #97, opened 2018-06-13
implement scrollToIndex()

### LINKED ISSUE #88: Programmatically scrolling to a given item/index
Looking at the main readme, there doesn't seem to be a way to programmatically scroll to a given item.

This is very useful in some contexts.

valdrinkoshi:
Fixes #88 by providing `scrollToIndex()` method, documentation and demo. This works both with all layouts we ship.

@domenic feel free to update the README.md as you see fit.

We invalidate the `_scrollToIndex` only when receiving scroll position changes (in other words, when user scrolls).

This implementation differs from #64 in the following: we don't compute the scroll anchor when user scrolls, nor allow customizing anchoring via itemAnchor and viewportAnchor.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -209,7 +209,7 @@ layout.scrollToIndex(9, 'center');
 layout.scrollToIndex(19, 'end');
 
 // Scroll to the 100th item, position it at the end of the viewport 
-// if we are scrolled above it already, otherwisef position it to the start.
+// if we are scrolled above it already, otherwise position it to the start.
 layout.scrollToIndex(99, 'nearest');
 
 ```
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -186,13 +186,32 @@ layout.updateItemSizes({
 });
 ```
 
-Set `layout.viewportScroll` to move the range across the container size.
+### Move range
+
+Use `viewportScroll (type: {top: number, left: number})` to move the range to a specific point.
 ```js
 const el = document.scrollingElement;
 el.addEventListener('scroll', () => {
   layout.viewportScroll = {top: el.scrollTop};
   layout.reflowIfNeeded();
 });
+```
+
+Use `scrollToIndex(index: number, position: string)` to move the range to a specific index.
+```js
+// Scroll to the 3rd item, position it at the start of the viewport.
+layout.scrollToIndex(2);
+
+// Scroll to the 10th item, position it at the center of the viewport.
+layout.scrollToIndex(9, 'center');
+
+// Scroll to the 20th item, position it at the end of the viewport.
+layout.scrollToIndex(19, 'end');
+
+// Scroll to the 100th item, position it at the end of the viewport 
+// if we are scrolled above it already, otherwisef position it to the start.
+layout.scrollToIndex(99, 'nearest');
+
 ```
 
 ## VirtualScroller (RepeatsAndScrolls mixin)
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -97,6 +97,17 @@ This re-renders all of the currently-displayed elements, updating them from thei
 
 This generally needs to be called any time the data to be displayed changes. This includes additions, removals, and modifications to the data. See our [examples below](#data-manipulation-using-itemschanged) for more information.
 
+### `scrollToIndex(index: number, position: string = "start")` method
```
valdrinkoshi: maybe rename `position` to `viewportPosition`? @domenic wdyt?

kenchris: It feels a bit like the `align` properties for flexbox. The position here also depends on the scroll direction.

Maybe you want to change the names of the values instead? `position: "viewport-start"` or `position: "scroll-direction-start"`

demo/scrolling.html
```diff
@@ -0,0 +1,86 @@
...
+    const virtualScroller = document.getElementById('scroller');
+
+    virtualScroller.updateElement = (element, contact, index) =>
+      element.textContent = `${index}) ${contact.longText}`;
+
+    fetch('contacts/contacts.json')
+      .then((resp) => resp.json())
+      .then((contacts) => virtualScroller.itemSource = contacts);
+
+    window.scrollToIndex = function(index, position) {
```
kenchris: Any reason why you mix arrow functions and non arrow functions?

README.md
kenchris: I see that `scrollIntoView` uses  `"start", "center", "end", or "nearest"` so let's keep that

> alignToTop: If true, the top of the element will be aligned to the top of the visible area of the scrollable ancestor. Corresponds to scrollIntoViewOptions: {block: "start", inline: "nearest"}. This is the default value.

Here the platform is using `block` and `inline` which is more a what than `position`. It seems a bit more future proof to create a `scrollToIndexOptions` dictionary

demo/scrolling.html
valdrinkoshi: not really, i can change this to arrow function 👌

README.md
domenic: +1 on using an options dictionary. I didn't realize that there was also a "behavior" member (hidden in https://drafts.csswg.org/cssom-view/#dictdef-scrolloptions) which we may want to add in the future.

That still leaves the question of what to call it. I think "position" is pretty good though. "Scroll to index 5 at the end position" sounds OK when said out loud.

(Working on updating this now.)

DESIGN.md
```diff
@@ -195,6 +197,23 @@ el.addEventListener('scroll', () => {
...
+
+// Scroll to the 10th item, position it at the center of the viewport.
+layout.scrollToIndex(9, 'center');
+
+// Scroll to the 20th item, position it at the end of the viewport.
+layout.scrollToIndex(19, 'end');
+
+// Scroll to the 100th item, position it at the end of the viewport 
+// if we are scrolled above it already, otherwisef position it to the start.
+layout.scrollToIndex(99, 'nearest');
```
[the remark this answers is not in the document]
valdrinkoshi: abruptly for now, i'm not sure how to implement smooth scrolling...probably we should move this method into VirtualScroller, which could check if it has the dom for the desired index, and call the native `scrollIntoView`..!

valdrinkoshi:
I'll merge this as is, we can add "behavior" implementation in a separate PR

Issue #101, opened 2018-06-13
Consider making the custom-element-with-template pattern even easier


domenic:
What follows is an idea that I don't think we should do right now, but we should write down and have a place to discuss, and maybe do eventually if the demand comes up a lot.

Background: [the section on using `<template>`s to automatically create the element](https://github.com/valdrinkoshi/virtual-scroller/tree/8301ce0330e04b93ea309e8832607b0fb63f6ed4#customizing-element-creation-and-updating-with-template)

The proposal is to make the custom-element-based pattern there easier. So instead of writing

```html
<virtual-scroller>
  <template>
    <contact-element sortable></contact-element>
  </template>
</virtual-scroller>

<script type="module">
  scroller.updateElement = (child, contact) => {
    child.contact = contact;
  };

  scroller.itemSource = contacts;
</script>
```

you would write something like

```html
<virtual-scroller elementproperty="contact">
  <template>
    <contact-element sortable></contact-element>
  </template>
</virtual-scroller>

<script type="module">
  scroller.itemSource = contacts;
</script>
```

Or maybe we'd just pick a default, and say that your custom element must have a property named `data` or `item` or something.

The default `updateElement` would then check to see if the appropriate data-recieving property existed, and if so, set it to the item from the ItemSource. If the property did not exist (e.g. because you were using a normal, non-custom, element), it would fall back to the current behavior of setting `textContent`.

This is fairly sugary, and just saves some typing. But if we see this pattern become one of the dominant ways to use virtual-scroller, we should consider adding it.

bennypowers:
Picking a default would be fine, but please retain the optional `elementproperty` (or some such) attribute. There will be uses where a ready-made custom element is available, but it doesn't take the default (sensible, though arbitrary) property.

rakina:
Closing this as this is related to the `virtual-scroller` prototype implementation and we intend to make this repository focus on the standards side of it.

PR #103, opened 2018-06-13
Expand on <template>-based element creation

### LINKED ISSUE #98: Explain relation between <template> default createElement and template instantiation proposal
(Note to self, will send PR soon)

domenic:
* Fix a few typos in the code samples.
* Add a pointer to #101 for anyone interested.
* Explain the relation with the template instantiation proposal, and our thoughts on future compatibility. Fixes #98.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ By setting a custom `updateElement` behavior, you can leverage more interesting 
   <template>
     <section>
       <h1></h1>
-      <img></img>
+      <img>
       <p></p>
     </section>
   </template>
@@ -171,7 +171,7 @@ By setting a custom `updateElement` behavior, you can leverage more interesting 
 
 <script type="module">
   scroller.updateElement = (child, contact) => {
-    child.querySelector("h1") = contact.name;
+    child.querySelector("h1").textContent = contact.name;
     child.querySelector("img").src = contact.avatarURL;
     child.querySelector("p").textContent = contact.bio;
   };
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -198,8 +198,36 @@ A useful pattern here is to encapsulate the details of updating your elements in
...
 
+#### Relation to template instantiation proposal
+
+The [template instantiation](https://github.com/w3c/webcomponents/blob/gh-pages/proposals/Template-Instantiation.md) proposal is a still-evolving idea for how to provide native updating of `<template>` elements to the web platform. This seems like exactly the kind of thing we'd want to use with `<virtual-scroller>`, once it becomes ready. For example, one could replace the above example with
+
+```html
+<virtual-scroller id="scroller">
+  <template>
+    <section>
+      <h1>{{name}}</h1>
```
valdrinkoshi: I'd update this example to use `{{contact.name}}` - which stresses another controversial point that the default processor needs to tackle, that is what context to pass into the template, e.g. 

```js
window.avatarBaseURL = 'http://my-cdn.com/contacts/id/';

scroller.updateElement = (section, contact, index) => {
  section.querySelector('h1').textContent = index + ') ' + contact.name;
  section.querySelector('img').src = avatarBaseURL + contact.id;
};
```
With template instantiation I'd expect to use:
```html
<virtual-scroller id="scroller">
  <template>
    <section>
      <h1>{{index}}) {{contact.name}}</h1>
      <img src="{{avatarBaseURL}}{{contact.id}}">
```
^ I'm not sure how the template instantiation would be able to get index and avatarBaseURL.
The virtual-scroller would have to pass the `contact` and `index` to the context of the template processor.

domenic: After thinking about this, I think we should not change the example to show more controversial default template processor features, as it kind of distracts from the point. Instead, we can assume that the item source always contains all the necessary information. So for your example, I might do something like

```js
scroller.itemSource = new ItemSource({
  item(index) {
    return {
      index,
      avatarURL: avatarBaseURL + contacts[index].id,
      name: contacts[index].name
    };
  },
  getLength() {
    return contacts.length;
  }
});
```

I wonder if we should add some helper for this in the future... e.g. ItemSource.fromArrayWithMappingFunction() or something. Maybe I'll file a tracking issue?

valdrinkoshi: Sounds good

PR #107, opened 2018-06-14
Use a content wrapper for applying transform to

kenchris:
Also encapsulate the overflow-x: hidden in the shadow root

### DIFF
--- a/demo/dismissable/dismissable-item.js
+++ b/demo/dismissable/dismissable-item.js
@@ -31,6 +31,19 @@ class DismissableItem extends HTMLElement {
   constructor() {
     super();
 
+    this.attachShadow({mode: 'open'}).innerHTML = `
+      <style>
+        :host {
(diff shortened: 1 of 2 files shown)


valdrinkoshi:
Nice! Small suggestion: I'd set `overflow: hidden` instead of `overflow-x: hidden`, so that in case we update this element in the future to support different dismiss directions (e.g. to be used inside a horizontal virtual-scroller), this part would be already good to go.
Also, maybe set it as an inline style on the `#contentWrapper` when we switch the state to `dragging`, and reset it to null when the gesture is done. This would completely hide this implementation detail from the outside world

kenchris:
I originally set it on the `#contentWrapper` instead of on `:host` but then I still saw horizontal scrollbar while dragging.

valdrinkoshi:
oh right - ofc 🤦‍♂️ 
LGTM! merging this

Issue #33, opened 2018-04-25
Requirements checklist


valdrinkoshi:
Tracker for the features to be implemented - copied from https://github.com/domenic/infinite-list-study-group/edit/master/REQUIREMENTS.md.

Mark complete after providing implementation and documentation.

## P0
- [x] Accurate scrollbar position for elements that have been loaded
  - Implemented, see [demo/layout.html](https://github.com/valdrinkoshi/virtual-list/blob/master/demo/layout.html) (scroll, change items count, and observe scroll position is kept)
- [ ] Overall list structure supports screen readers and accessibility controls
- [x] Heterogeneous elements
  - Controlled by the user via [`newChild` property](https://github.com/valdrinkoshi/virtual-list/blob/master/README.md#newchild-property)
- [x] Fallback element UI
  - Controlled by the user via [`newChild` property](https://github.com/valdrinkoshi/virtual-list/blob/master/README.md#newchild-property)
  - see [demo/fallback.html](https://github.com/valdrinkoshi/virtual-list/blob/master/demo/fallback.html)
- [x] Grid layout
  - `<virtual-list layout="vertical-grid"></virtual-list>`
  - see [`layout` property](https://github.com/valdrinkoshi/virtual-list/blob/master/README.md#layout-property)
- [x] List headers/footers
  - Controlled by the user via [`newChild` property](https://github.com/valdrinkoshi/virtual-list/blob/master/README.md#newchild-property)
  - see [demo/header-footer.html](https://github.com/valdrinkoshi/virtual-list/blob/master/demo/header-footer.html)
- [x] Configurable layout and style for elements in the list
  - User can style content via `virtual-list > *` or by adding arbitrary classes to the stamped children
  - layout configurable via the [`layout` property](https://github.com/valdrinkoshi/virtual-list/blob/master/README.md#layout-property)
- [x] Clickable list elements
  - Controlled by the user via `newChild/updateChild` properties
  - example in the README.md
- [ ] indexability of not loaded content 

## P1
- [x] Find in page (ctl+F) highlights elements that have been loaded
- [x] Sub groups within the list (i.e. letter groupings for a contact list)
  - see [demo/header-footer.html](https://github.com/valdrinkoshi/virtual-list/blob/master/demo/header-footer.html)
- [ ] Maintain scroll position across page load
- [x] "Load more" button
  - see [demo/loadmore.html](https://github.com/valdrinkoshi/virtual-list/blob/master/demo/loadmore.html)
[...]

PR #109, opened 2018-06-15
Add loadmore example

valdrinkoshi:
Addresses #33 - P1 "load more" example
 
loadmore

### DIFF
--- a/demo/loadmore.html
+++ b/demo/loadmore.html
@@ -41,15 +41,15 @@
 
     const button = document.createElement('button');
     const loadMore = async () => {
-      // Make loading go, then trigger itemsChanged so we can update
-      // the button status.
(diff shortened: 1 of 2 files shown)


demo/loadmore.html
```diff
@@ -0,0 +1,88 @@
...
+
+    const dataSource = new ContactDataSource();
+    const virtualScroller = document.querySelector('virtual-scroller');
+
+    const button = document.createElement('button');
+    const loadMore = async () => {
+      // Start loading the contacts...
+      const promise = dataSource.getContacts(20);
+      // ... notify list so the button can be updated...
+      virtualScroller.itemsChanged();
```
domenic: Would it work to move the updating code (setting disabled/textContent) to here? This feels a bit indirect.

PR #110, opened 2018-06-15
remove identity check

### LINKED ISSUE #93: should the itemSource getter return the original Array or the converted ItemSource?
virtual-scroller allows to set an Array as item source, and it will convert it to an ItemSource instance via ItemSource.fromArray().
The itemSource getter returns the converted instance instead of the original array. This might cause difficulties in checking for identity.
Should we:
1) keep it as is: accept Array in the setter, return ItemSource in the getter 
1) accept Array in the setter and return same Array in the getter, while using the generated ItemSource internally
1) accept only ItemSource instances in the virtual-scroller setter

I think 2) would work best

valdrinkoshi:
Fixes #93.

### DIFF
--- a/virtual-scroller-element.js
+++ b/virtual-scroller-element.js
@@ -11,7 +11,6 @@ const _createElement = Symbol();
 const _updateElement = Symbol();
 const _recycleElement = Symbol();
 const _nodePool = Symbol();
-const _rawItemSource = Symbol();
 const _itemSource = Symbol();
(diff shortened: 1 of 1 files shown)


domenic:
Hmm, so, I'm still not quite on board with this, sorry. I should have discussed more in #93. Let me try to do so, although I'm heading home soon...

valdrinkoshi:
@domenic updated to avoid doing identity checks 👌

PR #112, opened 2018-06-19
Consistently use .mjs to denote modules

mathiasbynens:
Module scripts behave differently than classic scripts. As such, I'd like to suggest explicitly denoting the difference by using the .mjs extension for modules and .js for classic scripts.

More details on this guidance: https://developers.google.com/web/fundamentals/primers/modules#mjs

Similar changes have landed upstream in the Layered APIs repository. It would be nice to be consistent with that.


mathiasbynens:
I've updated the commit message and the description, and have also rebased the patch. Please take another look. Thanks!

valdrinkoshi:
@mathiasbynens sorry I had to revert this as it breaks the `npm run format` and requires the server to handle `.mjs` MIME types - hence we should update the README.md.
Mind re-opening this PR with the fixes?

mathiasbynens:
Done: https://github.com/valdrinkoshi/virtual-scroller/pull/126 I’ve added a separate commit that changes `package*.json` and the README.

Issue #54, opened 2018-05-03
use IntersectionObserver?


valdrinkoshi:
`VirtualList` computes the viewport manually by diffing the scrollTarget bounds with the container bounds. We use `getBoundingClientRect()` to get scroll position and size, so we force layout.
We might be able to use [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) to get the scroll position info w/o causing relayout.
For the size, we can rely on ResizeObserver (#16)

domenic:
Is this still the plan? I think @flackr's virtual list was based on IntersectionObserver, right?

domenic:
The virtual-content branch is currently doing this :)

PR #119, opened 2018-06-27
Implement <virtual-content>

### LINKED ISSUE #113: Implement <virtual-content>
In https://github.com/valdrinkoshi/virtual-scroller/issues/72 we talked about `virtual-scroller` and `virtual-content`. This issue tracks the definition & implementation of `virtual-content`.

`<virtual-content>` auto-sizes itself according to the estimated size of the items (sets its min-width/height), and creates DOM only for the visible portion of it. It computes the portion of the viewport currently visible by listening for scroll events on the scroll parent. It can be contained in any scroll container.

valdrinkoshi:
Fixes #113.
We expose a private property to be able to override the scrollTarget when extending the custom element.

Note that we search for the scroll target on first attached, and that relies on the fact that the parent is already setup to be overflowing. It might be worth exposing a new property for `<virtual-content>` to set the scroll target, but let's wait to see if this is actually needed by collecting requirements and use cases first.


domenic:
I'll try to add some docs for this. Preferably without renaming the entire repository... ^_^

The scroll-target-finder is a bit tricky, and I wonder if there's any better way to do that, but nothing comes to mind immediately...

valdrinkoshi:
We could also expose it as a property and have its default be the `parentNode`.

valdrinkoshi:
Actually, if we use InteractionObserver #54 we shouldn't need to know who's the scroll target...For that reason I would not expose `scrollTarget` and keep it an implementation detail for now

domenic:
Closing this as this is related to the virtual-scroller prototype implementation and we intend to make this repository focus on the standards side of it. Thanks as always for your trailblazing work @valdrinkoshi! I think we did eventually come around to the point of view that "virtual content" style behavior is the right starting point.

PR #126, opened 2018-07-18
Consistently use .mjs to denote modules

mathiasbynens:
Module scripts behave differently than classic scripts. As such, I'd like to suggest explicitly denoting the difference by using the .mjs extension for modules and .js for classic scripts.

More details on this guidance: https://developers.google.com/web/fundamentals/primers/modules#mjs

Similar changes have landed upstream in the Layered APIs repository. It would be nice to be consistent with that.


domenic:
Readme line endings changed?

mathiasbynens:
They were CRLF for some reason. The patch makes them LF which is consistent with the other files.

domenic:
Can you separate that out into a separate commit, so that we can see the actual change?

mathiasbynens:
@domenic Done. (I ran `unix2dos README.md` and amended the commit.)

domenic:
Things keep changing in this repository, so I'll need to retarget this against the two new branches (see #155). Let me close this and commit directly to those, with you as the author.

PR #135, opened 2018-10-11
Formatting / spacing; removes unused properties, commented code; etc.

bicknellr:
To get adjusted to the code, I've been going through trying to find things to clean up. There are a few actual code changes in addition to the formatting / spacing stuff:
- Repeater:
  - Removed `_manageDom` and `_maintainDomOrder`, which aren't modified anywhere in the repo, and `_measuringId`, which was never referenced.
  - Setters that previously did nothing if you assigned the wrong type of value now throw.
  - Simplified some loop indexing stuff in `_reset`.
- Scroller:
  - The `container` getter is now `return super.container;` instead of `return this._container;`.
- Layout1D:
  - Removed `_maxIdx`, which was always `_totalItems - 1`.

### DIFF
--- a/virtual-repeater.js
+++ b/virtual-repeater.js
@@ -112,7 +112,6 @@ export const Repeats = Superclass => class extends Superclass {
   get first() {
     return this._first;
   }
-
   set first(idx) {
     if (typeof idx !== 'number') {
       throw new Error('New value must be a number.');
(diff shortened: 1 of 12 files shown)


virtual-repeater.js
```diff
@@ -344,10 +339,8 @@ export const Repeats = Superclass => class extends Superclass {
     const end = this._last;
     for (let idx = start; idx <= end; idx++) {
       const child = this._assignChild(idx);
-      if (this._manageDom) {
-        if (this._maintainDomOrder || !this._childIsAttached(child)) {
-          this._insertBefore(child, null);
-        }
+      if (!this._childIsAttached(child)) {
```
valdrinkoshi: argh, here we should have removed the if check completely in order to keep the dom order

PR #136, opened 2018-10-16
Rename some files; change mixins to classes.

bicknellr:
- Moves non-test, non-demo source into `/src/`.
- Rename classes based on their file name or imported name. (i.e. layouts are no longer all called `Layout` inside their file)
- Files centered around a single class now have the same name as that class. (except `VirtualScrollerElement` in `virtual-scroller.js`)
- Remove use of default exports.
- Mixins that were only used through classes made by mixing them on top of an empty class are now just classes.

Notably, this doesn't change any of the demo implementation that uses the built-in `std:virtual-scroller` copy, even though that demo would need to change if this were to replace it. I'll have to ask more about the expected workflow around that in the next meeting.


src/virtual-scroller.js
```diff
@@ -1,9 +1,7 @@
-import {_item, _key, ItemSource} from './item-source.js';
-import {default as Layout1dGrid} from './layouts/layout-1d-grid.js';
-import {default as Layout1d} from './layouts/layout-1d.js';
-import {VirtualScroller} from './virtual-scroller.js';
-
-export {ItemSource};
```
valdrinkoshi: why not exporting here?

bicknellr: It seemed odd that `ItemSource` was threaded through this file when other pieces like layouts weren't, so I figured it would be more consistent to say that all pieces you want should imported explicitly.

PR #138, opened 2018-11-01
Use node's http-server instead of python's SimpleHTTPServer.

fergald:
SimpleHTTPServer is single-threaded and often blocks for 20s on
keep-alive connections from chrome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -543,7 +543,8 @@ npm dependencies and serve from the project root
 
 ```sh
 $ npm install
-$ node_modules/http-server/bin/http-server -p 8081
+$ npm install -g http-server   // If http-server is not installed.
+$ http-server -p 8081
 ```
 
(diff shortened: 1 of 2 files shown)


fergald:
The python server causes 20-second network hangs with chrome. I spent time on debugging that until I found out that's it intrinsic to SimpleHTTPServer. It makes it impossible to usefully compare performance of virtualized vs non-virtualized pages. I really don't think it should be mentioned at all.

I have no strong preference for what should be mentioned but since this project already requires node, that seemed like low-effort choice. If there's a python builtin server that doesn't cause hangs, that's fine too.

fergald:
LGTM

valdrinkoshi:
Awesome! Mind updating this PR?

fergald:
Updated, PTAL.

PR #141, opened 2018-11-15
Integrate with lit-html 0.13

### LINKED ISSUE #134: Demo projects using lit-html broken
Hi,

After downloading and setting up the projects, all lit-html demos throw: 

screen shot 2018-09-10 at 07 04 30

lit-html version: v0.11.1

valdrinkoshi:
Fixes #134 by integrating to lit-html 0.13.0
Also, maintain dom order (regression introduced in #135)

### DIFF
--- a/virtual-repeater.js
+++ b/virtual-repeater.js
@@ -321,9 +321,8 @@ export const Repeats = Superclass => class extends Superclass {
     const end = Math.min(this._last, this._prevFirst - 1);
     for (let idx = end; idx >= start; idx--) {
       const child = this._assignChild(idx);
-      if (!this._childIsAttached(child)) {
-        this._insertBefore(child, this._firstChild);
-      }
+      // Maintain dom order.
+      this._insertBefore(child, this._firstChild);
       if (this.updateElement) {
         this.updateElement(child, idx);
       }
@@ -339,9 +338,8 @@ export const Repeats = Superclass => class extends Superclass {
     const end = this._last;
     for (let idx = start; idx <= end; idx++) {
       const child = this._assignChild(idx);
-      if (!this._childIsAttached(child)) {
-        this._insertBefore(child, null);
-      }
+      // Maintain dom order.
+      this._insertBefore(child, null);
       if (this.updateElement) {
         this.updateElement(child, idx);
       }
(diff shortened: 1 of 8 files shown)


virtual-repeater.js
```diff
@@ -321,9 +321,8 @@ export const Repeats = Superclass => class extends Superclass {
     const end = Math.min(this._last, this._prevFirst - 1);
     for (let idx = end; idx >= start; idx--) {
       const child = this._assignChild(idx);
-      if (!this._childIsAttached(child)) {
```
bicknellr: I think this line is important for the searchable + invisible DOM case. Is there something different with lit-html that requires this?

valdrinkoshi: This guarantees the dom order is kept. It's important also for accessibility (e.g. tabbing through elements follows dom order) and consistent stacking context, as we're positioning all children with transform + position absolute, and we want the next siblings to render above their predecessors.

valdrinkoshi: example where not doing so breaks a11y & stacking context: user scrolls down, then scrolls back up. When scrolling back up:
1. we collect and recycle elements from the tail
2. we reuse these elements for the head

If we don't keep the dom order, the user would have weird order while tabbing through the items

bicknellr: I think you're right about this being necessary for the truly virtualized (i.e. with recycling) case. I also think this breaks the searchable + invisible DOM case by always moving nodes that are being added to the visible range to either immediately before the last rendered range (here, using `this._insertBefore(child, this._firstChild);`) or the very end of the container (L342, using `this._insertBefore(child, null);`): it orders the rendered nodes correctly relative to each other but doesn't pay attention to anything that's not rendered and doesn't provide a way to prevent them from being moved, afaict. (The searchable + invisible DOM branches I had going relied on the virtual-scroller not moving nodes that were children of it.) Given that I haven't merged the searchable + invisible DOM branches, I still think this should go in, but I'm pretty sure we'll have to change it to fit that use.

valdrinkoshi: I see, then WDYT of restoring `_maintainDomOrder` property that was removed in #135? It could be configured in the constructor to false for the searchable + invisible DOM case

bicknellr: Yeah, that might be a reasonable way to add that feature back. I wonder if that flag should be exposed to the user too. I'm going to go ahead and merge this given that it's more correct given the current state of the repo and shouldn't be much work to add back when needed.

PR #152, opened 2019-01-22
<virtual-content>

bicknellr:
The virtual-content element is a different approach to implementing virtual list. This branch assumes that searchable-invisible DOM / display locking is available and necessary for allowing the content of the virtual list to be searchable, work with same-page anchor navigation, etc. This assumption means that all content that the user of the virtual-content is aware of should be inserted into the virtual-content element, which will manage the visibility of that content automatically. Additionally, because all content that the user of the virtual-content is aware of must be in the tree, this means that most of the original API isn't needed by virtual-content. (Concepts like element recycling and lazy item fetch / instantiation should be implemented on top of virtual-content.) See https://github.com/valdrinkoshi/virtual-scroller/issues/147 for more info.

For this PR:
- [x] Update README.md
- [x] Update DESIGN.md
- [x] Remove current demos broken by this PR.

For later PRs?
- [ ] Add new demos covering the same topics as those in master.

This PR is only meant for code review of the current content of the branch, without merging it into master (a separate PR).


kenchris:
Would be nice with some update to the readme or so, as this is removing all the existing implementation. Also, I assume that means that most of the demos will be broken.

domenic:
So, the intent here is to actually have two side-by-side branches, master and `virtual-content-2`, with virtual-content-2 prototyping a new approach based on [searchable invisible DOM](https://github.com/rakina/searchable-invisible-dom). We only opened this pull request because we want to do some initial code review of the virtual-content-2 branch; we won't actually merge it and replace the master branch.

Eventually, the virtual-content-2 branch may have the same capabilities and demos as the master branch, at which point it can replace it. But that's some ways off.

src/VirtualContent.js
```diff
@@ -0,0 +1,341 @@
...
+
+    const estimatedHeights = this[_estimatedHeights];
+    const updateHeightEstimate = (child) => {
+      if (isRenderable && !child.hasAttribute('invisible')) {
+        const childClientRect = child.getBoundingClientRect();
+        const style = window.getComputedStyle(child);
+        const height =
+          window.parseFloat(style.marginTop, 10) +
+          window.parseFloat(style.marginBottom, 10) +
+          childClientRect.height;
```
domenic: Can you just use `clientHeight`?

src/VirtualContent.js
```diff
@@ -0,0 +1,341 @@
...
+    );
+
+    const estimatedHeights = this[_estimatedHeights];
+    const updateHeightEstimate = (child) => {
+      if (isRenderable && !child.hasAttribute('invisible')) {
+        const childClientRect = child.getBoundingClientRect();
+        const style = window.getComputedStyle(child);
+        const height =
+          window.parseFloat(style.marginTop, 10) +
+          window.parseFloat(style.marginBottom, 10) +
```
domenic: It might be nicer to use typed OM: https://developers.google.com/web/updates/2018/03/cssom

src/VirtualContent.js
```diff
@@ -0,0 +1,341 @@
...
+    // renderable state (e.g. disconnected, invisible, `display: none`, etc.).
+    const isRenderable = !(
+      thisClientRect.top === 0 &&
+      thisClientRect.left === 0 &&
+      thisClientRect.width === 0 &&
+      thisClientRect.height === 0
+    );
+
+    const estimatedHeights = this[_estimatedHeights];
+    const updateHeightEstimate = (child) => {
```
domenic: This seems misnamed, since it returns a value. Perhaps `updateAndGetHeightEstimate()`.

src/VirtualContent.js
```diff
@@ -0,0 +1,341 @@
...
+      }
+      return estimatedHeights.get(child);
+    };
+
+    const previouslyVisible = new Set();
+    for (let child = this.firstChild; child !== null; child = child.nextSibling) {
+      if (!child.hasAttribute('invisible')) {
+        previouslyVisible.add(child);
+      }
+    }
```
domenic: I think a simpler version would be

```js
const previouslyVisible = new Set(this.querySelector('> *:not([invisible])'));
```

But maybe this is faster.

src/VirtualContent.js
```diff
@@ -0,0 +1,341 @@
...
+    this[_update]({forceVisible: new Set([child])});
+  }
+
+  [_scheduleUpdate]() {
+    if (this[_updateRAFToken] !== undefined) return;
+
+    this[_updateRAFToken] = window.requestAnimationFrame(this[_update]);
+  }
+
+  [_update]({forceVisible = new Set()} = {}) {
```
domenic: This method is pretty enormous. Is it possible to split it up into other methods? https://refactoring.guru/smells/long-method

For example, the closures you create could be methods that take extra arguments.

In general I wouldn't worry to much about symbol-naming everything; private methods should ship in V8 soon-ish, so underscore-prefixed is good enough for now.

src/VirtualContent.js
```diff
@@ -0,0 +1,341 @@
...
+
+          if (beforePreviouslyVisible) {
+            const scrollingAncestor = nearestScrollingAncestor(this);
+            if (scrollingAncestor !== null) {
+              scrollingAncestor.scrollBy(0, estimatedHeight - lastEstimatedHeight);
+            }
+          }
+        }
+
+        const isInViewport =
```
domenic: A comment should explain the difference between isInViewport and maybeInViewport, especially since they are computed in an identical way.

src/VirtualContent.js
domenic: Answer: no, they are different.

src/VirtualContent.js
domenic: Not going to do that for now because typed OM gives computed style, not resolved style, so we won't always get pixels.

src/VirtualContent.js
domenic: Going to leave as a to-do.

src/VirtualContent.js
```diff
@@ -0,0 +1,341 @@
...
+const DEFAULT_HEIGHT_ESTIMATE = 100;
+const TEMPLATE = `
+<style>
+:host {
+  /* Use flex to prevent children's margins from collapsing. Avoiding margin
+   * collapsing is simpler and good enough to start with. */
+  display: flex;
+  flex-direction: column;
+
+  /* Prevent the automatic scrolling between writes and later measurements,
```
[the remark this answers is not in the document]
bicknellr: "writes" here just refers to DOM modifications. Basically, without `overflow-anchor: none`, I noticed that the browser would automatically change the scroll position after some DOM edits. However, I'm trying to manually make those adjustments myself in `_update`, so I needed to prevent this behavior so that they wouldn't conflict.

src/VirtualContent.js
```diff
@@ -0,0 +1,341 @@
...
+        addedNodes: Array.from(this.childNodes),
+        removedNodes: [],
+        previousSibling: null,
+        nextSibling: null,
+      }
+    ]);
+    this[_mutationObserver].observe(this, {childList: true});
+
+    this.addEventListener(
+        'activateinvisible', this[_onActivateinvisible], {capture: true});
```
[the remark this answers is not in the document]
bicknellr: This listener is capturing to support the nesting case. If the listeners on nested `virtual-content` elements trigger from leaf to root, then they all are unrenderable until the last one removes `invisible` from the common ancestor. Capturing just makes sure they will remove the `invisible` attribute from root to leaf, but that's about all the nesting support I have in here.

src/VirtualContent.js
```diff
@@ -0,0 +1,341 @@
...
+          this[_resizeObserver].unobserve(node);
+          this[_intersectionObserver].unobserve(node);
+          node.removeAttribute('invisible');
+          estimatedHeights.delete(node);
+        }
+      }
+
+      for (const node of record.addedNodes) {
+        if (node.nodeType === Node.ELEMENT_NODE) {
+          // Added children should be invisible initially.
```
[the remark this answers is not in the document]
bicknellr: This function schedules an update at the end and that update will end up either accounting for the estimated space taken by the element in an empty space sentinels and the placement of the rendered children or by actually making the element visible. `_update` handles all of the visibility changes based on layout but newly added elements are initially made invisible here so that you can insert lots of elements without them all being rendered on the next frame.

src/VirtualContent.js
bicknellr: Unfortunately, relative selectors aren't supported by `querySelector{,All}` afaik, but I don't there's any good reason why I didn't at least try for something similar. I think I just didn't consider it since I was already iterating below. Maybe `Array.from(this.querySelectorAll('*:not([invisible])')).filter(node => node.parentNode === this)` would be worth trying out? Or adding some random garbage attribute for the sake of adding it to the selector so that the filtering bit can be dropped?

src/VirtualContent.js
bicknellr: When `maybeInViewport` is computed, `estimatedHeight` is either the current rendered height of the element or the last known (possibly inaccurate) height of the element, depending on whether or not the element was visible as of the last update. If `maybeInViewport` is true, then the element is made visible, its height is read, and the estimate* is updated again. The height estimate is now exact, and `isInViewport` is computed knowing the current, correct position of the element when rendered. If `isInViewport` is false, the element is made invisible and the true height of the element is added to the height of one of the empty space sentinel elements.

* Estimate is not a great choice of wording, there's no knowledgeable guessing involved.

domenic:
Review complete! Going to move this to a new clean `virtual-content` branch with a single commit, then link to that branch from `master`. Closing this PR!

PR #156, opened 2019-04-04
Make README an explainer

rakina:
Apologies in advance - there's more TODOs than content :) cc @domenic
Feel free to add to/change various parts, I'll work more on it tomorrow.


Prototypes.md
```diff
@@ -0,0 +1,25 @@
+## Prototypes
```
domenic: I think it's OK to just delete this text. It's not the focus of this repo. We won't delete the branches since we'll have them around for the historical record, but we don't need a top-level file explaining them.

README.md
```diff
@@ -12,28 +12,77 @@ We believe that, like native platforms, the web deserves a first-class virtual s
...
-But, there is much more work to do to show that this approach is feasible. We need to re-build the demos corpus to attain parity with the `traditional-virtualization` branch, to show that all the desired user experiences and framework integrations are still possible. And we need to work out an easy-to-use solution for populating the `<virtual-content>` element's initial set of children, without causing jank.
-
-## Talk at Chrome Dev Summit 2018
-
-At Chrome Dev Summit 2018, Gray Norton presented on our work so far. Give it a watch!
-
-[![virtual-scroller at Chrome Dev Summit 2018](http://img.youtube.com/vi/UtD41bn6kJ0/0.jpg)](http://www.youtube.com/watch?v=UtD41bn6kJ0)
+## Sample code
+
+```html
```
domenic: This example is great. I will push some tweaks but it does a great job illustrating the API in the way I was envisioning.

domenic:
@rakina I finally got a chance to work on this. Your review/fixes/additions would be appreciated!

I ran out of time before doing the "alternatives considered" section, but I think that's the only major piece left.

README.md
```diff
@@ -12,28 +12,185 @@ We believe that, like native platforms, the web deserves a first-class virtual s
...
+## Sample code
 
-We are not yet near the point of standardization, and are first working on proving out what the result should look like via JavaScript prototypes. So far this repository hosts two prototype implementations, each in separate branches:
+```html
+<!--
+  virtual-scroller lives in a built-in module and needs to be imported before use.
+  (The name of the module is subject to change.)
+-->
+<script type="module">
+import "std:virtual-scroller";
```
rakina: Is the different name of module vs element intentional?

README.md
```diff
@@ -12,28 +12,185 @@ We believe that, like native platforms, the web deserves a first-class virtual s
...
+### `rangeFirst` and `rangeLast` getters
+
+These return 0-based indices giving the first and last children currently rendered.
+
+### Constraints
+
+Ideally we would like there to be zero constraints on the contents of the `<virtual-content>` element, or on the virtual-content element itself.
+
+This may prove difficult to specify or implement, so we may need to add constraints such as:
+
```
rakina: Should we maybe also mention the "locked state"? Is that implementation detail?

README.md
```diff
@@ -12,28 +12,185 @@ We believe that, like native platforms, the web deserves a first-class virtual s
...
+* Support infinite-grid (spreadsheet-like) or masonry layouts.
 
-At Chrome Dev Summit 2018, Gray Norton presented on our work so far. Give it a watch!
+## Proposed APIs
 
-[![virtual-scroller at Chrome Dev Summit 2018](http://img.youtube.com/vi/UtD41bn6kJ0/0.jpg)](http://www.youtube.com/watch?v=UtD41bn6kJ0)
+### `<virtual-content>` element
+
+The `<virtual-content>` element represents a container that will manage the rendering of its children.
+The children of this element might not get rendered/updated if they are not near or in the viewport.
```
rakina: As a consequence of display-locking, the layout-related values might be observably different for virtualized vs non-virtualized children. Should we mention that?

README.md
domenic: I'm not sure :-/. Maybe we should just do `std:virtual-content` for now to keep them in sync, and figure out the final name (for both) later. That would mean our repository name, as well as the "concept" we talk about, is out of sync with the element name. But maybe that's not too bad.

Or, maybe we should just rename virtual-content to virtual-scroller, and use virtual-scroller everywhere for now, with the understanding that there's a final rename coming eventually. What do you think?

README.md
domenic: Yeah, that will be web-exposed, good point. I guess it's not much of a "constraint"... I'll see what I can do.

README.md
domenic: You mean, e.g., all children-of-children would have 0 width/height according to getBoundingClientRect? I will mention that in the new section I am adding. But I might have misunderstood, so please correct my text if so.

domenic:
Your text looks great, thank you! And thanks also for your fixes to my writing. I added the final section, and expanded the "Constraints" section into "Constraints and effects" per your comments. And then I did another tweaking pass.

To me this looks ready to merge, although the conflict between virtual-scroller and virtual-content naming is starting to feel a bit weird. Maybe we could resolve that first, then merge? But I'm eager to see this land either way.

README.md
rakina: Not exactly, it will have approximately correct width and height and offsets etc, which might be different than the actual values when they are made visible, because of the placeholder sizes. Changed the wording to "Changing the layout of non-visible children's descendants", is that to broad?

domenic: That seems perfect!

rakina:
Thanks for the additions! Renamed everything to `virtual-scroller`. Also just realized I don't actually have write access. Can you merge this?

domenic:
Oh, eek! @valdrinkoshi would you mind giving @rakina write access? We will work on getting this repo moved to WICG soon enough so it's not a hassle for you any more :)

valdrinkoshi:
Done @rakina you should have received the invite for collaboration 👌

PR #170, opened 2019-08-15
add instructions for trialling with chrome

fergald:


fergald:
Merged this since domenic is OOO, pls review after you're back.

PR #171, opened 2019-08-21
make and use confirm-features custom element

fergald:
Turn the code of confirm-features.mjs into a custom element to make it trivial to reuse in all demos.

### DIFF
--- a/demo/loadmore.html
+++ b/demo/loadmore.html
@@ -28,6 +28,7 @@
       height: 80px;
     }
   </style>
+  <script type="module" src="./util/confirm-features.mjs"></script>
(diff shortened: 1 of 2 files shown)


demo/loadmore.html
```diff
@@ -45,7 +45,6 @@
   </script>
   <script type="module">
```
domenic: This should probably be moved to the head (it seems nicer to import things before they're used, just for code readability) and then can be simplified to `<script type="module" src="./util/confirm-features.mjs"></script>`

fergald: done.

PR #172, opened 2019-08-21
Fergald simple demos

fergald:
Add several simple demos. 

Also adds some custom elements that are share among these and other demos.

### DIFF
--- a/demo/simple/index.html
+++ b/demo/simple/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
-  <title>Virtual Scoller Demos</title>
+  <title>Virtual Scoller Simple Demos</title>
 
   <p>Simple demos that show FPS performance for some common actions. In
     all cases you can switch between a virtual scroller element and a
--- a/demo/simple/index.html
+++ b/demo/simple/index.html
@@ -1,0 +1,22 @@
+<p>Simple demos that show FPS performance for some common actions. In
+all cases you can switch between a virtual scroller element and a
+plain div as the container. You can also adjust the number of elements
+in the container.
+
+<ul>
+  <li>
+    <a href="resize.html">Resize the scroller</a>
+  </li>
+  <li>
+    <a href="resize-font.html">Resize the font of the scroller content</a>
+  </li>
+  <li>
+    <a href="restyle.html">Change the style of the scroller content</a>
+  </li>
+  <li>
+    <a href="scroll.html">Smoothly scroll through the content</a>
+  </li>
+  <li>
+    <a href="move-first-child.html">Move content around</a>
+  </li>
+</ul>
--- a/demo/simple/index.html
+++ b/demo/simple/index.html
@@ -1,22 +1,23 @@
-<p>Simple demos that show FPS performance for some common actions. In
-all cases you can switch between a virtual scroller element and a
-plain div as the container. You can also adjust the number of elements
-in the container.
+<!DOCTYPE html>
+<html lang="en">
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <title>Virtual Scoller Demos</title>
 
-<ul>
-  <li>
-    <a href="resize.html">Resize the scroller</a>
-  </li>
-  <li>
-    <a href="resize-font.html">Resize the font of the scroller content</a>
-  </li>
-  <li>
-    <a href="restyle.html">Change the style of the scroller content</a>
-  </li>
-  <li>
-    <a href="scroll.html">Smoothly scroll through the content</a>
-  </li>
-  <li>
-    <a href="move-first-child.html">Move content around</a>
-  </li>
-</ul>
+  <p>Simple demos that show FPS performance for some common actions. In
+    all cases you can switch between a virtual scroller element and a
+    plain div as the container. You can also adjust the number of elements
+    in the container.
+
+  <ul>
+    <li>
+      <a href="resize.html">Resize the scroller</a>
+    <li>
+      <a href="resize-font.html">Resize the font of the scroller content</a>
+    <li>
+      <a href="restyle.html">Change the style of the scroller content</a>
+    <li>
+      <a href="scroll.html">Smoothly scroll through the content</a>
+    <li>
+      <a href="move-first-child.html">Move content around</a>
+  </ul>
--- a/demo/simple/move-first-child.html
+++ b/demo/simple/move-first-child.html
@@ -1,0 +1,50 @@
+<head>
+  <title>Virtual Scoller Demo - Moving children</title>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, minimum-scale=1.0, initial-scale=1, user-scalable=yes">
+  <style>
+    .scroller {
+        overflow: scroll;
+        height: 400px;
+    }
+  </style>
+</head>
+
+<body>
+  <p>Repeatedly moving the first child to the end of the content. This
+  is potentially faster as we may be able to avoid layout changes to
+  the locked elements, however the scroller's internal state must
+  react to the mutation.</p>
+
+  <script type="module">
+    // Do this separately. If virtual-scroller is not available, the demo
+    // will still work.
+    import 'std:elements/virtual-scroller';
+  </script>
+  <script type="module">
+    import '../util/confirm-features.mjs';
+  </script>
+  <script type="module">
+    import '../util/fps-display.mjs';
+  </script>
+  <script type="module">
+    import '../util/demo-controller.mjs';
+  </script>
+  <confirm-features></confirm-features>
+  <fps-display></fps-display>
+
+  <demo-controller id=dc></demo-controller>
+  <div style="width: 100%">
+    <div id=contentDiv class="scroller resize">
+      <virtual-scroller id=vs></virtual-scroller>
+    </div>
+  </div>
+
+  <script type="module">
+    import * as Util from '../util/util.mjs';
+    dc.setContainer(vs);
+    Util.everyFrame(i => {
+    vs.appendChild(vs.firstChild);
+    });
+  </script>
+</body>
--- a/demo/simple/move-first-child.html
+++ b/demo/simple/move-first-child.html
@@ -9,26 +9,13 @@
         height: 400px;
     }
   </style>
+  <script type="module" src="./scaffolding.mjs"></script>
 
   <p>Repeatedly moving the first child to the end of the content. This
     is potentially faster as we may be able to avoid layout changes to
     the locked elements, however the scroller's internal state must
     react to the mutation.</p>
 
-  <script type="module">
-    // Do this separately. If virtual-scroller is not available, the demo
-    // will still work.
-    import 'std:elements/virtual-scroller';
-  </script>
-  <script type="module">
-    import '../util/confirm-features.mjs';
-  </script>
-  <script type="module">
-    import '../util/fps-display.mjs';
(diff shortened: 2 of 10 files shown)


demo/simple/index.html
```diff
@@ -0,0 +1,22 @@
+<p>Simple demos that show FPS performance for some common actions. In
```
domenic: It'd be better to include the preamble:

```html
<!DOCTYPE html>
<html lang="en">
<meta charset="utf-8">
<meta name="viewport" content="width=device-width, initial-scale=1">
<title>Simple virtual-scroller demos</title>
```

This is mostly best-practice stuff that won't matter in this particular case:

- Avoids quirks mode
- Declare to screenreaders, scrapers, etc. that you're using English
- Avoids any content being interpreted as the legacy default charset (windows-1252)
- Makes it appear reasonable on mobile
- Makes the tab have a reasonable title

demo/simple/move-first-child.html
```diff
@@ -0,0 +1,50 @@
+<head>
```
domenic: Missing some things from the preamble. (Also I think the viewport contains some redundant stuff.)

demo/simple/move-first-child.html
```diff
@@ -0,0 +1,50 @@
...
+  <meta name="viewport" content="width=device-width, minimum-scale=1.0, initial-scale=1, user-scalable=yes">
+  <style>
+    .scroller {
+        overflow: scroll;
+        height: 400px;
+    }
+  </style>
+</head>
+
+<body>
```
domenic: I don't care too much but I prefer the minimal style of no head/body tags ([and so does the Google style guide](https://google.github.io/styleguide/htmlcssguide.html#Optional_Tags)). Whereas the html tag is necessary as a place to attach the lang=en.

demo/simple/move-first-child.html
```diff
@@ -0,0 +1,50 @@
...
+  is potentially faster as we may be able to avoid layout changes to
+  the locked elements, however the scroller's internal state must
+  react to the mutation.</p>
+
+  <script type="module">
+    // Do this separately. If virtual-scroller is not available, the demo
+    // will still work.
+    import 'std:elements/virtual-scroller';
+  </script>
+  <script type="module">
```
domenic: All of these could be a bit nicer by just doing `<script type="module" src="..."></script>`

domenic: However you may want to just have a single module that's something like demo-scaffolding.mjs which:

- Imports all three of these
- Adds a confirm-features element
- Adds a fps-display element
- Adds a demo-controller element
- Dynamic imports `import("std:elements-virtual-scroller")` (dynamic import failure won't prevent subsequent code from executing)

That could really DRY up the demos.

demo/simple/move-first-child.html
```diff
@@ -0,0 +1,50 @@
...
+    <div id=contentDiv class="scroller resize">
+      <virtual-scroller id=vs></virtual-scroller>
+    </div>
+  </div>
+
+  <script type="module">
+    import * as Util from '../util/util.mjs';
+    dc.setContainer(vs);
+    Util.everyFrame(i => {
+    vs.appendChild(vs.firstChild);
```
domenic: Indentation seems off here

demo/simple/move-first-child.html
```diff
@@ -0,0 +1,50 @@
...
+  <demo-controller id=dc></demo-controller>
+  <div style="width: 100%">
+    <div id=contentDiv class="scroller resize">
+      <virtual-scroller id=vs></virtual-scroller>
+    </div>
+  </div>
+
+  <script type="module">
+    import * as Util from '../util/util.mjs';
+    dc.setContainer(vs);
```
domenic: It's not great practice to rely on the ID -> global mapping, and I'd kind of prefer not to spread it it in demos. So just adding `const dc = document.getElementById("dc")` (or the querySelector equivalent) would be nice.

demo/simple/resize-font.html
```diff
@@ -0,0 +1,55 @@
...
+
+  <demo-controller id=dc></demo-controller>
+  <div style="width: 100%">
+    <div id=contentDiv class="scroller resize">
+      <virtual-scroller id=vs></virtual-scroller>
+    </div>
+  </div>
+
+  <script type="module">
+    dc.setContainer(vs);
```
domenic: This `dc.setContainer(vs)` could also go in a demo-scaffolding.mjs file. (Note: modules always run at the end of the page by default so there's no worry about needing to split the code into before the HTML/after the HTML.)

demo/simple/scroll.html
```diff
@@ -0,0 +1,50 @@
...
+  <style>
+    .scroller {
+        overflow: scroll;
+        height: 400px;
+    }
+  </style>
+</head>
+
+<body>
+  <p>Scroll a little every frame. This will be slower as the virtual
```
domenic: "but hopefully not by much", maybe?

demo/util/demo-controller.mjs
```diff
@@ -0,0 +1,124 @@
+import * as Util from './util.mjs';
+
+/**
+ * Create a string of |wordCount} words.
```
domenic: } -> |

demo/util/demo-controller.mjs
```diff
@@ -0,0 +1,124 @@
...
+  let result = '';
+  for (let i = 0; i < wordCount; i++) {
+    result += ' word';
+  }
+  return result;
+}
+
+const NUM_WORDS = 50;
+const WORDS = words(NUM_WORDS);
+const DEFAULT_COUNT = 10000;
```
domenic: default count of what?

demo/util/demo-controller.mjs
```diff
@@ -0,0 +1,124 @@
...
+    <button class=count>10000</button>
+    <button class=count>100000</button>
+  </div>
+  <button id=swap></button>
+</div>
+`;
+
+/**
+ * Displays buttons that change the number of elements in a container
+ * and also whether the container is a vitual-scroller or a plain
```
domenic: vitual -> virtual

demo/util/demo-controller.mjs
```diff
@@ -0,0 +1,124 @@
...
+    this.#container = container;
+    this.#updateContainer();
+  }
+
+  #setCount = count => {
+    this.#count = count;
+    this.#updateContainer();
+  }
+
+  #swapTo = () => {
```
domenic: Naming implies it'll actually swap; maybe getSwappedLocalName or something

demo/util/fps-display.mjs
```diff
@@ -0,0 +1,51 @@
...
+  #fps
+
+  constructor() {
+    super();
+
+    const shadowRoot = this.attachShadow({mode: 'closed'});
+    shadowRoot.innerHTML = TEMPLATE;
+    this.#fps = shadowRoot.getElementById('fps')
+      .getElementsByClassName('place');
+    this.#scheduleUpdate();
```
domenic: It seems you could use everyFrame (and remove scheduleUpdate) if you modified everyFrame to pass in the timestamp as well as the index. That might also remove #counter.

demo/simple/move-first-child.html
fergald: all fixed and consistent across files.

demo/simple/index.html
fergald: Done.

demo/simple/move-first-child.html
fergald: Done.

demo/simple/move-first-child.html
fergald: Done. Didn't know about dynamic import.

demo/simple/move-first-child.html
fergald: fixed.

demo/simple/move-first-child.html
fergald: done and move into scaffolding.mjs

demo/simple/resize-font.html
fergald: done.

demo/simple/scroll.html
fergald: Add "but hopefully fast enough to avoid dropping a frame."

demo/util/demo-controller.mjs
fergald: done

demo/util/demo-controller.mjs
fergald: done

demo/util/demo-controller.mjs
fergald: done

demo/util/demo-controller.mjs
fergald: done.

demo/util/fps-display.mjs
fergald: That's much nicer.

fergald:
Rebased and addressed all comments.

demo/simple/index.html
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
+<html lang="en">
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <title>Virtual Scoller Demos</title>
```
domenic: Maybe "Simple Demos" to be distinct from https://github.com/WICG/virtual-scroller/pull/173

demo/simple/restyle.html
```diff
@@ -0,0 +1,41 @@
...
+  <demo-controller id=dc></demo-controller>
+  <div style="width: 100%">
+    <div id=contentDiv class="scroller resize">
+      <virtual-scroller id=vs></virtual-scroller>
+    </div>
+  </div>
+
+  <script type="module">
+    import * as Util from '../util/util.mjs';
+    let divStyle;
```
domenic: I think this could be written as

```js
const divStyle = [...document.styleSheets[0].rules].find(r => r.selectorText === "#vs > div").style;
```

demo/util/fps-display.mjs
```diff
@@ -0,0 +1,47 @@
...
+`;
+
+const MS_PER_SECOND = 1000;
+
+/**
+ * Displays frames-per-second performance of the page for recent frames.
+ */
+export class FpsDisplay extends HTMLElement {
+  #lastTime = null;
+  #fps
```
domenic: Missing semicolon

demo/util/fps-display.mjs
```diff
@@ -0,0 +1,47 @@
...
+      .getElementsByClassName('place');
+    Util.everyFrame((n, timestamp) => {
+      this.#update(n, timestamp);
+    });
+  }
+
+  #update = (n, timestamp) => {
+    if (this.#lastTime !== null) {
+      const delta = timestamp - this.#lastTime;
+      this.#fps[n % this.#fps.length].innerText =
```
domenic:
```suggestion
      this.#fps[n % this.#fps.length].textContent =
```

demo/util/util.mjs
```diff
@@ -16,3 +16,22 @@ export function swapElement(swapOut, swapIn) {
...
+  let i = 0;
+  function update(timestamp) {
+    callback(i, timestamp);
+    schedule();
+    i++;
+  }
+  function schedule() {
+    window.requestAnimationFrame(
+      update
+    );
```
domenic: Can't this fit on one line?

demo/util/fps-display.mjs
fergald: Fixed. Odd that eslint doesn't find this. I think it's similar to the other rule I fixed, which has yet to be pulled into head, so I guess I'll leave this.

PR #174, opened 2019-08-22
add automatic load more

fergald:
Very basic, just replaced the load-more button with a DIV and uses default intersection observer to spot when it has been scrolled into view. I decided against using a buffer on the intersection observer to preload the content because I want it to be clearly visible what is going on.

### DIFF
--- a/demo/loadmore.html
+++ b/demo/loadmore.html
@@ -85,7 +85,9 @@
       }
     };
 
-    const observer = new IntersectionObserver(loadMore, {});
+    // Will trigger when the first pixel of the loader becomes
+    // visible.
+    const observer = new IntersectionObserver(loadMore);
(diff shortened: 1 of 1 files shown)


demo/loadmore.html
```diff
@@ -51,24 +51,31 @@
...
         let contactElement = document.createElement('contact-element');
-        button.before(contactElement);
+        loader.before(contactElement);
         contactElement.contact = contact;
       }
     };
-    button.addEventListener('click', loadMore);
-    loadMore();
+
+    const observer = new IntersectionObserver(loadMore, {});
```
domenic:
```suggestion
    const observer = new IntersectionObserver(loadMore);
```

fergald: done

PR #176, opened 2019-08-23
add .clang-format file and reformat all .mjs files

fergald:
This is chromium's .clang-format file.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,39 @@
+# Defines the Chromium style for automatic reformatting.
+# http://clang.llvm.org/docs/ClangFormatStyleOptions.html
+BasedOnStyle: Chromium
+# This defaults to 'Auto'. Explicitly set it for a while, so that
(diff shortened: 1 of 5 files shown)


.clang-format
```diff
@@ -0,0 +1,39 @@
+# Defines the Chromium style for automatic reformatting.
```
domenic: I don't think we should just copy the .clang-format file from Chromium, as these demos are not part of Chromium and won't become so. I think it'd be better to have a minimal file (which seems like it'd just be the first 3 lines?).

fergald: done

CodeWitchBella:
Why not use prettier for js formatting?

fergald:
Simply because we're using clang-format elsewhere. If prettier is the preferred formatter for wicg stuff I'm fine with that.

@domenic is it appropriate to drop a `package.json` in here to get prettier?

PR #178, opened 2019-08-28
Build a map of contacts

fergald:
From the suggestion in https://github.com/WICG/virtual-scroller/pull/177/commits/c136e46b5a74129eae3e06289661c2491af55f04

This constructs a map of letter->[contact].


fergald:
Agree that this doesn't improve things.

PR #179, opened 2019-08-30
add performance demo mode

fergald:
Add a button that enables an FPS display and animates resizing the contact list horizontally. Between virtual-scroller and raw divs demonstrates clear drop in FPS and increase in CPU.

### DIFF
--- a/demo/sticky.html
+++ b/demo/sticky.html
@@ -35,6 +35,20 @@
         opacity: 0.9;
         z-index: 1;
     }
+    .resize {
+        animation-direction: alternate;
+        animation-duration: 1s;
+        animation-iteration-count: infinite;
+        animation-timing-function: linear;
+        animation-name: resize;
+    }
+    @keyframes resize {
+        0% { width: 400px; }
+        100% { width: 600px; }
+    }
+    .scroller {
+        overflow: scroll;
+    }
   </style>
 </head>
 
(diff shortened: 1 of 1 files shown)


demo/sticky.html
```diff
@@ -54,12 +68,16 @@
 
   <div>
     <button id=swapButton>switch virtual-scroller &lt;-&gt; div</button>
+    <button id=perfDemoModeButton>performance demo mode</button>
   </div>
 
-  <virtual-scroller id=scroller></virtual-scroller>
-
+  <fps-display style="display: none"></fps-display>
```
domenic: Personally I would use the `hidden` attribute (plus `toggleAttribute("hidden", perfModeDemoState)`)

demo/sticky.html
```diff
@@ -131,6 +149,32 @@ <h1>${firstLetter}</h1>
...
+    let perfDemoModeState = false;
+
+    /**
+     * Swaps between performance demo mode and not. In this mode we
+     * should the FPS and we do an animated resize of the contacts list.
+     */
+    function perfDemoMode() {
+      let fpsDisplayStyleDisplay;
+      let scrollContainerClass;
+      if (perfDemoModeState === false) {
```
domenic: Do `if (perfDemoModeState) { ... } else { ... }`

demo/sticky.html
```diff
@@ -131,6 +149,32 @@ <h1>${firstLetter}</h1>
     }
     document.querySelector('#swapButton')
         .addEventListener('click', swapScrollers);
+
+    let perfDemoModeState = false;
```
domenic: Maybe `isInPerfDemoMode`

domenic:
I tried this on mobile and you can't see the resize. (However FPS still drop drastically with divs.) It might be worth tweaking the stylesheet a bit to make that work.

demo/sticky.html
fergald: yeah, no idea what I was doing there. Anyway the code is a lot simpler now.

demo/sticky.html
fergald: done.

demo/sticky.html
fergald: I started to do that but then that either gives me 2 indepdent bits for the state of whether we're in demo mode or not (one in JS, one in this attribute) or I can switch to make this the canonical bit but that feels weird.

But I did switch to toggling the hidden attribute.

fergald:
I'm going to submit this and look into fixing mobile after.

PR #181, opened 2019-08-30
Use switches in demo

domenic:
I thought this would be cool so I spent 10 minutes hacking it up on top of #179. It does look a lot nicer.

/cc @tkent-google

We probably shouldn't merge this because it just makes https://github.com/WICG/virtual-scroller/issues/180 worse. We could productionize it by creating a switch polyfill and using import maps.

If we were to move forward we should probably also use switches in the perf/ demos folder.

### DIFF
--- a/demo/sticky.html
+++ b/demo/sticky.html
@@ -16,6 +16,11 @@
       flex-direction: column;
     }
 
+    label {
+      display: block;
+      margin: 1em;
+    }
+
     virtual-scroller {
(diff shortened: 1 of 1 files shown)


fergald:
Yeah, it looks cool but as you say, needs a polyfill or something.

domenic:
Will maybe re-attempt this over at fergald/virtual-scroller-demos; closing for now.

PR #193, opened 2019-10-26
Update README.md: undefined variable scroller

Ffloriel:


README.md
```diff
@@ -38,6 +38,8 @@ import "std:virtual-scroller";
 </virtual-scroller>
 
 <script>
+const scroller = document.getElementById('scroller');
```
fergald: `scroller` should be a global referring to the element with that id. At least in chrome, there is no undefined variable error. What browser is giving you an error?

That said, `newChildren` below *is* undefined but this isn't really meant to be real functioning code anyway.

